PR #13, opened 2021-10-05
Update integration workflow

jtcohen6:
Partial resolution of dbt-labs/dbt#3924. If we agree with the approach here, we can extend to `dbt-snowflake` + `dbt-bigquery`.

- Run the `integration` workflow nightly at 5 UTC, to identify any regressions caused by changes in `dbt-core`.
- Install `dbt-core` from `HEAD` (instead of hardcoding `develop` branch) to get latest changes
- Remove logic from `integration` workflow that picks plugins to test based on file changes. This is nice for PRs that only edit the README, but it's not as important now that this repo only contains the Redshift plugin. This logic seemed to be running into bugginess on `push` triggers ([e.g.](https://github.com/dbt-labs/dbt-redshift/runs/3777004562)); we'll be able to see after merging this PR.

### DIFF
--- a/.github/scripts/integration-test-matrix.js
+++ b/.github/scripts/integration-test-matrix.js
@@ -8,47 +8,39 @@ module.exports = ({ context }) => {
     // `changes` is a list of adapter names that have related
     // file changes in the PR
     // ex: ['postgres', 'snowflake']
-    const changes = JSON.parse(process.env.CHANGES);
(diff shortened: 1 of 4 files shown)


jtcohen6:
Ah, it was actually the _unit_ tests that caught the breaking change (i.e. deprecation enacted) in https://github.com/dbt-labs/dbt/pull/3975. Nice!

Should we consider running both workflows nightly?

PR #35, opened 2021-11-05
Bump artifact versions in tests

jtcohen6:
This is a bad test to have in the adapter plugin, we should remove or change it in the long run. For now, manual updates.


jtcohen6:
A few tests are failing, here and in this morning's scheduled run, because of https://github.com/dbt-labs/dbt-redshift/issues/36. This fixes the _other_ tests that failed in the scheduled run, so it's still a net improvement :)

PR #42, opened 2021-11-18
 Add Redshift parameter to create tables with backup option specified

### LINKED ISSUE #18: Redshift create table with backup option
### Describe the feature
The `CREATE TABLE` statement in Redshift has a configuration option that controls whether or not the table should be included in automated and manual cluster snapshots.  The default setting for this option is `YES`.  At present this is not configurable in DBT and the default is used for all table and incremental materialisations.

By allowing configuration of this parameter in DBT models we can save processing time when creating snapshots and restoring from snapshots and to reduce storage space on Amazon Simple Storage Service.

There is no way to get around this except to turn off automated snapshots at a cluster level.

> To disable automated snapshots, set the retention period to zero. If you disable automated snapshots, Amazon Redshift stops taking snapshots and deletes any existing automated snapshots for the cluster.

### Additional context
This feature is Redshift specific.  The feature is similar but not exactly the same as Snowflakes `TRANSIENT` table.

### Who will this benefit?
This will benefit all DBT users who use Redshift and who want to control storage costs and to speed up snapshot creation and restores.

P.S. If you think this is worthwhile and in keeping with DBT's principles, then I would love to work on this issue with a PR.

dlb8685:
resolves #18

### Description

This PR adds a config parameter "backup" that controls whether a Redshift table will be created with `BACKUP No` or not. Current default (both in DBT and more generally with Redshift) is for this BACKUP to be set to True unless explicitly overridden. But, for some transient and intermediate tables, there can be no need for backup and users would like to be able to set config accordingly.

The config parameter is added in `impl.py` and the statements in `adapters.sql` are updated to use this parameter.

### Checklist

I don't have a Redshift instance to run this code on. I will investigate if the unit tests fail, but based on nearly identical old [PR](https://github.com/dbt-labs/dbt-core/pull/3221) and local linting I'm optimistic.


dlb8685:
@jtcohen6 I was hoping the tests would run in CI to prove this out a bit, but I'm a first-time contributor. I don't have a Redshift database to do much with local testing. Let me know your thoughts in general and if there's anything that needs to be added.

jtcohen6:
Closing and reopening to trigger adapter integration tests

tests/integration/backup_table_tests/test_backup_table_option.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def check_backup_param_template(self, test_table_name, backup_is_expected):
+        # Use raw DDL statement to confirm backup is set correctly on new table
+        with open('target/run/test/models/{}.sql'.format(test_table_name), 'r') as ddl_file:
+            ddl_statement = ddl_file.readlines()
+            self.assertEqual('backup no' not in ' '.join(ddl_statement).lower(), backup_is_expected)
+
+    @use_profile('redshift')
+    def test__redshift_backup_table_option(self):
+        self.assertEqual(len(self.run_dbt(["seed"])), 1)
```
jtcohen6, line 36: There aren't any seeds defined in the project
```suggestion
```

tests/integration/backup_table_tests/test_backup_table_option.py
```diff
@@ -0,0 +1,69 @@
...
+        self.check_backup_param_template('model_backup_false', False)
+
+        # Any view should not contain a BACKUP NO parameter, regardless of the specified config (create will fail)
+        self.check_backup_param_template('model_backup_true_view', True)
+
+    @use_profile('redshift')
+    def test__redshift_backup_table_option_project_config_false(self):
+        # Update project config to set backup to False.
+        # This should make the 'model_backup_undefined' switch to BACKUP NO
+        self.project_config_append('models', {'backup': False})
```
jtcohen6, line 55: Looks like this is causing an error in the integration test:
```
    def project_config_append(self, new_name, new_element):
>       self._project_config[new_name] = new_element
E       AttributeError: 'TestBackupTableOption' object has no attribute '_project_config'
```

Much simpler would be to create a separate test case, and define the config anew:
```python
TestBackupTableOptionProjectFalse(TestBackupTableOption):

    @property
    def project_config(self):
        return {
            'config-version': 2,
            'models': {'backup': False}
        }

    @use_profile('redshift')
    def test__redshift_backup_table_option_project_config_false(self):
        ...
```

jtcohen6 (commented):
Thanks for reopening the PR here @dlb8685!

Looks like the tests are failing in a few small ways, should be quick to fix up. We're cutting a second (and last) RC of v1.0 very soon, so there's a good chance this goes into v1.1 instead, which is totally ok :)

tests/integration/backup_table_tests/test_backup_table_option.py
dlb8685: I made this change and also removed the lines that were running the seed. Let me know if this looks good and if the tests can run again.

jtcohen6 (approved):
@dlb8685 Nice work! I think we might be able to sneak this in for v1.0.0rc2 after all :)

Just added you as a contributor in the changelog, will merge as soon as the tests pass.

dlb8685:
Thank you! @jtcohen6

will-peloton:
I reported a bug I think that is related to this #60. I believe the fix is moving backup option up 2 lines in the adapter.sql before table_attribute.

PR #43, opened 2021-11-18
Fire typed events in `tests.integration.base` instead of AdapterLogger

jtcohen6:
Nightly runs have been failing since https://github.com/dbt-labs/dbt-core/pull/4266 was merged:
- https://github.com/dbt-labs/dbt-redshift/actions/runs/1474953855
- https://github.com/dbt-labs/dbt-redshift/actions/runs/1470106194

In particular, a handful of tests that use `--vars` were failing here:
```python
tests/integration/simple_snapshot_test/test_simple_snapshot.py:95: in _run_snapshot_test
    self.run_dbt(['snapshot', '--vars', '{seed_name: seed_newcol}'])
tests/integration/base.py:409: in run_dbt
    res, success = self.run_dbt_and_check(args=args, profiles_dir=profiles_dir)
tests/integration/base.py:444: in run_dbt_and_check
    logger.info("Invoking dbt with {}".format(final_args))
```
Complaining about key errors in the `*args` being passed along:
```python
>       msg = self.base_msg.format(*self.args)
E       KeyError: 'seed_name'
```

I'm guessing this has to do with the lazy evaluation for log messages that we spent time pulling apart and understanding in https://github.com/dbt-labs/dbt-core/pull/4266. I'm just surprised this started failing _after_ we merged those fixes.

For the time being, I was able to get these tests passing again locally by making `tests/integration/base.py` look more like it does in dbt-core: fire typed events (`IntegrationTestInfo`, `IntegrationTestDebug`, `IntegrationTestException`), instead of plumbing through `AdapterLogger`.

We very badly need to (a) improve this integration testing scaffold, and (b) consolidate its logic into a single module (package?) that can be reused across core + all our adapters. For now, these need to be manual fixups.

### DIFF
--- a/tests/integration/base.py
+++ b/tests/integration/base.py
@@ -25,11 +25,14 @@ from dbt.logger import log_manager
 from dbt.events.functions import (
     capture_stdout_logs, fire_event, setup_event_logger, stop_capture_stdout_logs
 )
-from dbt.events import AdapterLogger
+from dbt.events.test_types import (
+    IntegrationTestInfo,
+    IntegrationTestDebug,
+    IntegrationTestException
+)
 from dbt.contracts.graph.manifest import Manifest
 
 
-logger = AdapterLogger("Redshift")
 INITIAL_ROOT = os.getcwd()
 
 
@@ -239,7 +242,7 @@ class DBTIntegrationTest(unittest.TestCase):
                 'test_original_source_path={0.test_original_source_path}',
                 'test_root_dir={0.test_root_dir}'
             )).format(self)
-            logger.exception(msg)
+            fire_event(IntegrationTestException(msg=msg))
 
             # if logging isn't set up, I still really want this message.
             print(msg)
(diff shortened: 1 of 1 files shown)


emmyoop:
Is this really only failing in integration tests?  Have we locally triggered log messages in adapters that use args?  Looks like Redshift only has one in `connections.py`.

nathaniel-may:
It looks like the underlying problem here is with lines like the following:
```
>>> "dict: {'k':'v'}".format()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
KeyError: "'k'"
```

Here's what I see happening:
1. redshift calls `logger.info("Invoking dbt with {}".format(final_args))`
2. dbt.events.types.py::AdapterEventBase::message calls `msg = self.base_msg.format(*self.args)` which triggers a `KeyError` like the example above does.

Since this could be triggered by any other log messages in the adapters, I suggest we use the solution in dbt-core#4305 instead. I'm not opposed to using test events here, but I think I'd like adapters to use one form of logging to make community contributions simpler to navigate.

jtcohen6:
Closing in favor of https://github.com/dbt-labs/dbt-core/pull/4305.

We should still aim for a future where all adapters can use the exact same integration testing scaffold as `dbt-core` + `dbt-postgres`, and where we don't have to coordinate small updates to `tests.integration.base` across several repos

PR #51, opened 2021-12-07
Fixing Action to post test failures

leahwicz:
The `notify_when` field doesn't act how I thought it was supposed to. It looks at failures in the current job but not other jobs upstream which is what we want.

Previous logic: Always run post job, but only post if failures
New logic: Only run post job on failures, but always post


McKnight-42 (approved):

McKnight-42:
code seems fine, but seeing its behind main

iknox-fa (approved):
LGTM

PR #63, opened 2022-01-11
Fix: Create table statement parameter ordering when configured with backup & dist/sort 

### LINKED ISSUE #60: [CT-23] syntax error with table backup option
### Describe the bug
dbt run fails on syntax error for models that have both backup option and table_attribute set e.g. dist_key

### Steps To Reproduce
run this model 

```
{{ config(backup=False, dist='a', sort='a') }}
select 1 as a
```

### Expected behavior
model should build

### Screenshots and log output
Screen Shot 2022-01-05 at 1 05 37 PM

### System information
**The output of `dbt --version`:**
```
installed version: 1.0.1
   latest version: 1.0.1

Up to date!

Plugins:
  - snowflake: 0.21.0
  - redshift: 1.0.0
  - postgres: 1.0.1
```

**The operating system you're using:**
Mac
**The output of `python --version`:**
Python 3.8.6
### Additional context
backup option needs to be before table_attribute, which is not the case by looking at complied SQL. This is related to https://github.com/dbt-labs/dbt-redshift/pull/42/files

SMeltser:
resolves #60 

### Description
This PR corrects the parameter ordering of the table creation statement to accommodate when a model is configured with both a BACKUP parameter as well as an accompanying DIST/SORT key. 

### Checklist


McKnight-42:
@SMeltser if you could update your local branch and repush up to test with most recent changes to main we should be able to get this approved.

VersusFacit:
👍 To what Matt said 

Since this was let through before because we didn't have a test for this case, would you mind adding a test or two for queries which include backup and various other Redshift options? 

Not to send you out into deep end by yourself. Here's the [class](https://github.com/dbt-labs/dbt-redshift/blob/e8297d58536d2f08b6159d51008df14126c9ff6b/tests/integration/backup_table_tests/test_backup_table_option.py#L6) or thereabouts a test has both the backup option and a dist key/sort key should go!

We'll be happy to lend extra support in making sure this behavior gets recorded as a regression test.

SMeltser:
Will do ASAP! Thank you for the feedback @McKnight-42 @VersusFacit

SMeltser:
@McKnight-42 @VersusFacit Just completed the test updates. Open to any and all feedback you have regarding the setup. I incorporated a conditional flow if the backup param is expected to ensure that the ordering of parameters is valid when dist/sortkey's are supplied.

tests/integration/backup_table_tests/test_backup_table_option.py
```diff
@@ -27,10 +27,16 @@ def check_backup_param_template(self, test_table_name, backup_is_expected):
         with open('target/run/test/models/{}.sql'.format(test_table_name), 'r') as ddl_file:
             ddl_statement = ddl_file.readlines()
             self.assertEqual('backup no' not in ' '.join(ddl_statement).lower(), backup_is_expected)
```
VersusFacit, line 29: I think I want this to be an exception that gets raised rather than an assertion since that'd be taken as a test....I think? Seems that way from the [docs](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertEqual) for UnitTest. A plain assert with a message explain use would be sufficient and not trigger a test failure but rather an outright error. Maybe that's overkill?

Do you have a opinion on which would be "best"?

tests/integration/backup_table_tests/test_backup_table_option.py
```diff
@@ -27,10 +27,16 @@ def check_backup_param_template(self, test_table_name, backup_is_expected):
...
+            if backup_is_expected:
+                distkey_index = ' '.join(ddl_statement).lower().find('distkey')
```
VersusFacit, line 31: Rather than reinvoking this 3, might as well store it in a variable and run the finds on that? Unless there's a specific state element at play, although I don't believe that's the case.

VersusFacit (changes requested):
Hey, this is great! Just two minor comments. I really like the trick of checking things by index in the string.

tests/integration/backup_table_tests/test_backup_table_option.py
SMeltser: @VersusFacit While im not extremely familiar with the original purpose of using a configurable flag for assertion, I think that it was to handle circumstances where the state of the project overall (via defaults in yml, etc) conflict or differ with the configuration of an individual model. It is definitely a unique approach If we go the exception route we would have to most likely change some of the test models accordingly.

will-peloton:
thank you all for working this fix. This may be a bit off topic but when can we expect 1.0.1 to be released? I'm really looking forward to using this.

tests/integration/backup_table_tests/test_backup_table_option.py
VersusFacit: Nope you're right. Sorry, we're making some changes to this testing framework and I'm still getting up to speed with (changing) best practices. Forgive the red herring.

VersusFacit:
@SMeltser We had a few bugs crop up over the last week or so that are responsible for unit tests failures in your logs. Thankfully, we've already fixed them. 

Merging the adapter root branch into this project will resolve them and get you green, at which point I'll (finally) approve for.

@will-peloton As soon as we get this green, I think we can merge it. As for 1.0.1 of this adapter (if that's what you mean; please correct me if I'm misreading), we're very much in a 'keeping the lights on' mode right now. That said, I'll report back to the team a desire to see that initiative pushed through.

VersusFacit (approved):
Nicely done!!! So glad to see this completed :)) Thanks for sticking with it 🙇‍♀️ 🚀

PR #81, opened 2022-03-11
Initial code to use dbt.tests.adapter in Redshift

### LINKED ISSUE #78: [CT-351] Use the new testing framework in the redshift plugin
As part of the testing overhaul we have created a suite of adapter tests that will be located in the dbt.tests.adapter namespace and will be available for subclassing in specific adapter test suites. This ticket is to enable using the new test suite and implement the initial batch of tests (the ones that replace dbt-adapter-tests).

gshank:
resolves #78 

### Description

Use the new testing framework in Redshift.

### Checklist

### DIFF
--- a/tests/functional/adapter/test_basic.py
+++ b/tests/functional/adapter/test_basic.py
@@ -1,0 +1,79 @@
+import pytest
+
+from dbt.tests.adapter.basic.test_base import BaseSimpleMaterializations
+from dbt.tests.adapter.basic.test_singular_tests import BaseSingularTests
+from dbt.tests.adapter.basic.test_singular_tests_ephemeral import (
+    BaseSingularTestsEphemeral,
+)
+from dbt.tests.adapter.basic.test_empty import BaseEmpty
+from dbt.tests.adapter.basic.test_ephemeral import BaseEphemeral
+from dbt.tests.adapter.basic.test_incremental import BaseIncremental
+from dbt.tests.adapter.basic.test_generic_tests import BaseGenericTests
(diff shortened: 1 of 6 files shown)


tests/functional/adapter/test_basic.py
```diff
@@ -0,0 +1,79 @@
...
+class TestIncrementalRedshift(BaseIncremental):
+    pass
+
+
+class TestGenericTestsRedshift(BaseGenericTests):
+    pass
+
+
+class TestSnapshotCheckColsRedshift(BaseSnapshotCheckCols):
+    # Redshift defines the 'name' column such that it's not big enough
```
ChenyuLInx, line 58: Should we consider updating the base case to accommodate all adapters? but can also leave this as future steps

ChenyuLInx (approved):
LGTM!

tests/functional/adapter/test_basic.py
gshank: We can't update the base case to make Redshift work, because it requires a specific column datatype which doesn't work for all adapters.

jtcohen6: I do think this is related to [some jank around updating](https://github.com/dbt-labs/dbt-core/blob/ea5a9da71ef71cc13414cd8c2f8097448add5edb/core/dbt/tests/util.py#L261-L327) these fixture tables mid-sequence, which is worth discussing further, alongside the jank we've already identified: https://github.com/dbt-labs/dbt-core/issues/4778#issuecomment-1075515724

PR #131, opened 2022-07-07
[CT-789] Add Grants to Materializations #212

### LINKED ISSUE #128: [CT-789] Add support for grant macros
Everything in Redshift is _mostly_ the same as in PostgreSQL, so we don't require any changes to:
- materializations
- `get_grant_sql`
- `get_revoke_sql`

However, the syntax for showing grants is _very_ different. I think something like this will work, though I'll be the first to admit that it looks & feels pretty weird:
```sql
{% macro redshift__get_show_grant_sql(relation) %}

with privileges as (

        -- valid options per https://docs.aws.amazon.com/redshift/latest/dg/r_HAS_TABLE_PRIVILEGE.html
	select 'select' as privilege
	union all
	select 'insert' as privilege
	union all
	select 'update' as privilege
	union all
	select 'delete' as privilege
	union all
	select 'references' as privilege

)

select
    u.usename as grantee,
    p.privilege
from pg_user u
cross join privileges p
where has_table_privilege(u.usename, '{{ relation }}', privilege)

{% endmacro %}
```
Even though that function is called `has_table_privilege`, it works for both tables + views. Schema-level privileges require using `has_schema_privilege` instead; not something we need to worry about for now.

Assuming that query is able to return columns by the same name (and we could rename them in the SQL if need be), we shouldn't need to override any other parts of the puzzle.

dbeatty10:
resolves #128

### Description

Includes the tests in:
- https://github.com/dbt-labs/dbt-core/issues/5437
- https://github.com/dbt-labs/dbt-core/pull/5447
- https://github.com/dbt-labs/dbt-core/pull/5452

### Todo before merge
- [x] Remove unused `DBT_TEST_USER_*` environment variables from [secrets](https://github.com/dbt-labs/dbt-redshift/settings/secrets/actions)
- [x] Update `dev-requirements.txt`

### Checklist


McKnight-42:
@cla-bot check

.github/workflows/integration.yml
```diff
@@ -170,6 +167,9 @@ jobs:
           REDSHIFT_TEST_USER: ${{ secrets.REDSHIFT_TEST_USER }}
           REDSHIFT_TEST_PORT: ${{ secrets.REDSHIFT_TEST_PORT }}
           REDSHIFT_TEST_HOST: ${{ secrets.REDSHIFT_TEST_HOST }}
+          DBT_TEST_USER_1: ${{ secrets.DBT_TEST_USER_1 }}
+          DBT_TEST_USER_2: ${{ secrets.DBT_TEST_USER_2 }}
+          DBT_TEST_USER_3: ${{ secrets.DBT_TEST_USER_3 }}
```
jtcohen6, line 172: We might not actually need these as repo secrets, versus just in plain text here

dbt/include/redshift/macros/adapters/apply_grants.sql
```diff
@@ -0,0 +1,25 @@
...
+    select 'references' as privilege_type
+
+)
+
+select
+    u.usename as grantee,
+    p.privilege_type
+from pg_user u
+cross join privileges p
+where has_table_privilege(u.usename, '{{ relation }}', privilege_type)
```
jtcohen6, line 23: I think we can also filter out grants to the current user here:
```sql
and u.usename != current_user
```

tests/functional/adapter/test_grants.py
```diff
@@ -0,0 +1,41 @@
...
+
+my_snapshot_sql = """
+{% snapshot my_snapshot %}
+    {{ config(
+        check_cols='all', unique_key='id', strategy='check',
+        target_database=database, target_schema=schema
+    ) }}
+    select 1 as id, cast('blue' as {{ type_string() }}) as color
+{% endsnapshot %}
+""".strip()
```
dbeatty10, line 17: Without this override, I was getting this error upon the 2nd of 2 consecutive snapshots:
```
failed to find conversion function from "unknown" to text
```

### Root cause
Old versions of Postgres (and apparently current versions of Redshift) have a known issue with detecting types of untyped constants [[1]](https://github.com/OHDSI/SqlRender/issues/66)[[2]](https://stackoverflow.com/questions/18073901/failed-to-find-conversion-function-from-unknown-to-text)[[3]](https://www.postgresql.org/message-id/65937bea0901060104n399cdec8ye5e2b8e247e5139a@mail.gmail.com)

dbeatty10: Optionally, we could add this (slightly) more verbose version to dbt-core.

On that note, is there syntax for type-casting that we could enable for core + adapters?

I'd prefer less verbose syntax like this:
```
select 1 as id, {{ cast_string('blue') }} as color
```
over this:
```
select 1 as id, cast('blue' as {{ type_string() }}) as color
```

The former would also allow some adapters to render more pithy SQL like:
```
'blue'::text
```
rather than:
```
cast('blue' as text)
```

.github/workflows/integration.yml
dbeatty10: Done

dbt/include/redshift/macros/adapters/apply_grants.sql
```diff
@@ -0,0 +1,27 @@
...
+    and u.usename != current_user
```
dbeatty10, line 24: It would be awesome if there is a test that fails if this line is missing.

Do you know if that is possible @jtcohen6?

tests/functional/adapter/test_grants.py
jtcohen6: @dbeatty10 I think we could add a cross-db `{{ cast() }}` macro to `dbt-core`, which returns `column::type` where it's supported, or `cast(column as type)` otherwise / by default.

For now, I did the simpler version in https://github.com/dbt-labs/dbt-core/pull/5452/commits/7f499ba2c6d71545bfab9023ff3d182d9b100c98

jtcohen6:
Rerunning CI to account for changes in dbt-labs/dbt-core@c763601. I don't believe any code changes here are needed.

dbt/include/redshift/macros/adapters/apply_grants.sql
dbeatty10: > @dbeatty10 I've confirmed that our existing tests actually DO fail if this line is missing.

Perfect!

tests/functional/adapter/test_grants.py
dbeatty10: Awesome, I'll try the relevant update and see if it works.

McKnight-42 (approved):
Great work @dbeatty10

PR #147, opened 2022-08-03
Avoid truncating Redshift model names with more than 63 characters

Goodkat:
resolves [#5586](https://github.com/dbt-labs/dbt-core/issues/5586)

### Description

Because the dbt-redshift adapter inherits from dbt-postgres, it will use any new dbt-postgres methods/macros unless told otherwise. However Redshift allows relation names up to 127 characters ([docs](https://docs.aws.amazon.com/redshift/latest/dg/r_names.html)), so it doesn't require the same complex handling at the 63-character mark as it is done for Postgres.

### Checklist


jtcohen6:
@Goodkat Responding to your comment in the issue:

> I am not sure about the tests, since I have no access to Redshift just now. Can we mock it?

We have access to a Redshift cluster running in CI. I think it would be simplest to add a functional test with the reproduction case from the original issue. In that example, the two models were named:
- `my_example__shipping_group_configurations__path_groups_and_settings__path_group__shipping_path_codes`
- `my_example__shipping_group_configurations__path_groups_and_settings`

Goodkat:
@jtcohen6 I found out the existing tests for Redshift model names "my_name_is_XX_characters...":
https://github.com/dbt-labs/dbt-redshift/tree/main/tests/integration/relation_name_tests/models

We can extend them with the new entries for 127 characters or add  there the cases from the original issue. What do you think?

Goodkat:
And who maintains the ```CHANGELOG.md``` for this repository? It doesn't support ```changie``` here, right?
How to update it properly?

troyharvey:
@Goodkat Example of how to update the `CHANGELOG` here: https://github.com/dbt-labs/dbt-redshift/commit/9a3492a1c3394496c9061252d54c87caa112821a#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed

VersusFacit:
Thanks for taking a crack at this @Goodkat. Do you feel confident writing a test for this? I see you've got a good look model for this but you'll need a Python driver. I also appreciate this is time sensitive.

Goodkat:
Thanks @VersusFacit for your comments. I just found and extended the existing test models with the new case for 127 characters. I am not sure if it is enough for testing this.

VersusFacit:
Hitting you back. I wanted  to add a negative test to make sure the exception handling worked -- it does -- and in my PR, I also added tests in our new framework. 

I'm closing this one out, but because I cherry picked your commits, your work will be rightfully highlighted in the history upon merge 💥

PR #295, opened 2023-01-31
convert backup table integration tests to functional tests

### LINKED ISSUE #293: [CT-1948] Convert backup_table_tests

### Current Behavior

backup_table_tests are run in the "integration" framework

### Expected Behavior

backup_table_tests are run in the "functional" framework

### Steps To Reproduce

1. run integration tests

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt-core:
- dbt-redshift:
```

### Additional Context

_No response_

mikealfare:
resolves #293

### Description

This PR converts the backup table integration tests to use the new "functional" tests framework.

### Checklist


.changes/unreleased/Under the Hood-20230130-205628.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Under the Hood
```
McKnight-42, line 1: nit: test conversions don't necessarily need a changelog entry, can apply skip changelog label in github if you wish instead

tests/functional/adapter/conftest.py
```diff
@@ -0,0 +1,24 @@
+import pytest
+
+
+@pytest.fixture
```
McKnight-42, line 4: Curious what the advantage of having this be in the conftest vs a fixture defined in file?

tests/functional/adapter/test_backup_table.py
```diff
@@ -0,0 +1,78 @@
+import pytest
```
McKnight-42, line 1: Would it be worth bringing in some of the comments from the previous tests back to explain whats going on in each test? wonder if any contributors who might want to modify or work in this area might find changing up the test to be complicated if they are more used to something like pandas or ae side of things

tests/functional/adapter/conftest.py
VersusFacit: Any chance of there being a conflict with existing `conftest.py`? I'm guessing "no" since you got all green on the test suite. But, I'm not confident as to how pytest decides to scope `conftest.py` files.

tests/functional/adapter/test_backup_table.py
```diff
@@ -0,0 +1,78 @@
...
+
+from dbt.tests.util import run_dbt
+
+
+_MODELS = {
```
VersusFacit, line 6: I really think -- as annoying as this is -- we  should have these be split into individual model variables and then included where necessary. Right now, it's hard to know what's needed where and the readability is lower. What happens if any of these extended the line past 100 chars for example? This may be needed on a new model. I'd rather we follow the existing convention.

Also myself wouldn't really care if you extracted the model definitions into their own variables and built `_MODELS` like you have here.

tests/functional/adapter/test_backup_table.py
```diff
@@ -0,0 +1,78 @@
...
+        yield run_dbt(["run"])
+
+    @pytest.fixture(scope="class")
+    def models(self):
+        return _MODELS
+
+
+class TestBackupTableSetup(BackupTableBase):
+
+    def test_setup_executed_correctly(self, run_dbt_results):
```
VersusFacit, line 29: Why have this as a test? Is it good to have tests test the test itself (lol, wording) or should that be an assert? Curious what you have to say.

tests/functional/adapter/test_backup_table.py
```diff
@@ -0,0 +1,78 @@
...
+    "syntax_with_distkey.sql": "{{ config(materialized='table', backup=False, dist='my_col') }} select 1 as my_col",
+    "syntax_with_sortkey.sql": "{{ config(materialized='table', backup=False, sort='my_col') }} select 1 as my_col",
+}
+
+
+class BackupTableBase:
+
+    @pytest.fixture(scope="class", autouse=True)
+    def run_dbt_results(self, project):
+        yield run_dbt(["run"])
```
VersusFacit, line 20: 1. `["run"]` isn't needed
2. `yield` is a super overloaded term in pytest so it'd be rad to have a quick comment explaining what you're doing with this function

tests/functional/adapter/test_backup_table.py
```diff
@@ -0,0 +1,78 @@
...
+
+    @pytest.mark.parametrize(
+        "model_ddl,search_phrase",
+        [
+            ("syntax_with_distkey", "diststyle key distkey"),
+            ("syntax_with_sortkey", "compound sortkey"),
+        ],
+        indirect=["model_ddl"]
+    )
+    def test_properly_formed_ddl(self, model_ddl, search_phrase):
```
VersusFacit, line 58: I'm not sure that's what this test is testing. More like "backup predicate comes before secondary predicates" or something like that. Right now, especially with how dense your test is, this is a bit tricky to read.

tests/functional/adapter/conftest.py
```diff
@@ -0,0 +1,24 @@
...
+        "model_ddl,backup_expected",
+        [("backup_is_false", False)],
+        indirect=["model_ddl"]
+    )
+    def test_setting_reflects_config_option(self, model_ddl: str, backup_expected: bool):
+        backup_will_occur = "backup no" not in model_ddl.lower()
+        assert backup_will_occur == backup_expected
+    ===
+
+    This example will return the contents of the backup_is_false DDL file, which can be referenced via model_ddl.
```
VersusFacit, line 21: Mind editing this comment to reference that `yield` returns the contents and does so as a string. Took me and Matt a bit to figure out what this was doing since you joined on `' '` instead of `\n'. It's functional, just a bit unexpected/unclear.

VersusFacit (changes requested):
Some great and model use of the `pytest` framework. Just think this one will benefit from some commenting and renaming of elements within for readability.

tests/functional/adapter/test_backup_table.py
```diff
@@ -0,0 +1,78 @@
...
+    def project_config_update(self):
+        return {"models": {"backup": False}}
+
+    @pytest.mark.parametrize(
+        "model_ddl,backup_expected",
+        [
+            ("backup_is_true", True),
+            ("backup_is_undefined", False)
+        ],
+        indirect=["model_ddl"]
```
VersusFacit, line 74: Classy as hell. I wish this feature was better documented.

tests/functional/adapter/conftest.py
mikealfare: Re: @McKnight-42, I wanted this to be reusable for other tests. If we don't need to read project files for anything else, then it might not be appropriate to put it here. But I assumed that checking the parsed sql was something that might be more common in an adapter test.

Re: @VersusFacit, Closer to the test takes precedent. You only expose the fixture to tests that are in files while are alongside that particular conftest.py or in files that are in folders/subfolders alongside that particular conftest.py.

tests/functional/adapter/conftest.py
mikealfare: Oh, fair. So would it make more sense to update the comment? Or to join on a newline? I can explicitly put string in the comment, but it's also in the type annotation. Let me know what makes the most intuitive sense and I'll make the change.

tests/functional/adapter/test_backup_table.py
mikealfare: I put this here because the integration test was effectively doing this for each test. So I removed that piece and ran it once instead of n times. It also helped me make sure that I was setting this up correctly as this was my first conversion. I agree it can be removed.

tests/functional/adapter/test_backup_table.py
mikealfare: I can add that in. Basically all of the tests just look for the presence (or lack thereof) of the string "backup no" in the resulting ddl to determine if the config made it through the pipe. One of them makes sure it's in the right spot (loosely).

tests/functional/adapter/test_backup_table.py
mikealfare: I didn't realize a module variable per model was the convention. I can make that change. I went with the more compact version simply because it fit, as you noted. That being said, I think they're all needed for the class, so I would just move this into `def models(self)` referencing the variable instead of the string literal. Also, running it at the class level makes it that I only need to run dbt once, which is the long pole in the tent.

tests/functional/adapter/test_backup_table.py
mikealfare: 1. I like being explicit, especially in tests
2. I actually don't think there's a difference between `yield` and `return` here since there's no teardown. I'm used to just using `yield` in this situation because of pytest. The only reason I need the results of `run_dbt()` is for the test below that we seemed to agree should be removed. So I'll remove that test and not return anything from the fixture.

tests/functional/adapter/test_backup_table.py
mikealfare: That's correct, and I struggled for a name. I'm fine changing the name as suggested.

tests/functional/adapter/conftest.py
mikealfare: Re: @McKnight-42 I completely misunderstood your comment. Yes, this can be in a separate fixtures file. There's no benefit to my knowledge. When I only have a few fixtures, I'll put them in conftest.py for simplicity. But once you hit a certain point, it's better to have it in a file. I could see us getting there, so we can start that here. I'll make the change.

mikealfare: @VersusFacit I found this while researching the best way to structure this: https://docs.pytest.org/en/6.2.x/fixture.html#:~:text=described%20here.-,conftest,-.py%3A%20sharing%20fixtures. It verifies how the scope works.

mikealfare: After going through everything, I decided to keep the fixture in `conftest.py`. There are multiple things that you can configure in the main `conftest.py`, but once you're in one level, the only purpose I can find for this file is to scope fixtures. In other words, this _is_ the fixtures file already.

tests/functional/adapter/test_backup_table.py
VersusFacit: Thanks for adding the newline here. For whatever reason, it just makes it much easier to picture what is happening (that you're reconstructing the file).

VersusFacit (approved):
LGTM. Thanks for explaining your reasoning and linking to docs. I'm learning the best practices for `pytest` too :)

.changes/unreleased/Under the Hood-20230130-205628.yaml
mikealfare: TIL

PR #309, opened 2023-02-03
CT-1960: test conversion - simple snapshot tests

### LINKED ISSUE #257: [CT-1776] Convert remaining integration tests (dbt-redshift)
The [integration folder](https://github.com/dbt-labs/dbt-redshift/tree/main/tests/integration) should be emptied, all remaining tests should be converted to the new framework in the [functional folder](https://github.com/dbt-labs/dbt-redshift/tree/main/tests/functional).

Most conversions should be ports from the core repo, others will be specific to Redshift.

mikealfare:
resolves #257

### Description

Converts snapshot tests to functional tests (from integration tests).
Includes a few new utility methods that may be reused in other redshift functional tests.

### Checklist


tests/functional/adapter/common.py
```diff
@@ -0,0 +1,136 @@
+from typing import Dict, Tuple, List
+
+from dbt.tests.util import relation_from_name
+from dbt.tests.fixtures.project import TestProjInfo
+
+
+def get_records(project: TestProjInfo, table: str, select: str = None, where: str = None) -> List[Tuple]:
```
colin-rogers-dbt, line 7: looks like these aren't redshift specific, should we just open a PR in core to add them?

tests/functional/adapter/test_snapshot.py
```diff
@@ -0,0 +1,204 @@
...
+    get_records,
+    update_records,
+    insert_records,
+    delete_records,
+    clone_table,
+    add_column
+)
+
+
+_SEED = """
```
colin-rogers-dbt, line 17: I think the rule we've been following is if these constants are > 50 lines we've been moving them to a separate fixture.py file

tests/functional/adapter/test_snapshot.py
```diff
@@ -0,0 +1,204 @@
...
+        unique_key='id',
+        strategy='check',
+        check_cols=['email'],
+    ) }}
+    select * from {{ ref('fact') }}
+{% endsnapshot %}
+"""
+
+
+class SnapshotBase:
```
colin-rogers-dbt, line 87: Looks like dbt-core has a snapshot test: https://github.com/dbt-labs/dbt-core/blob/main/tests/functional/simple_snapshot/test_basic_snapshot.py

Could we just move those tests to the adapter zone and inherit from them or if these are better move these into core?

colin-rogers-dbt (commented):
It seems like a lot of this code might actually be a better fit for living in dbt-core

tests/functional/adapter/common.py
mikealfare: Eventually. My approach is:
1) put them in dbt-redshift
2) get feedback on whether we want to make changes
3) open a PR to put them in dbt-core
4) open a PR to removed them from dbt-redshift and use the dbt-core versions

That way we're not waiting on PRs from two different repos to get this out or point to two different branches of dbt-core like we do in other tests.

tests/functional/adapter/test_snapshot.py
mikealfare: This is 31, but I can do that if you want.

tests/functional/adapter/test_snapshot.py
mikealfare: I don't want to inherit tests from another repo. I think it's a bad pattern.

- The foreign repo tests become too integrated with the functional code and you wind up with a polluted interface
- It leads to the scenario where you need to submit parallel PRs in multiple repos to update a test case
- It's not as easy to see what's actually being tested

tests/functional/adapter/test_snapshot.py
colin-rogers-dbt: Was including all the constants, up to you whether it's better for readability

tests/functional/adapter/test_snapshot.py
colin-rogers-dbt: Re-evaluating our pattern for test inheritance/management is probably out of scope for just converting our tests. Not saying that the current pattern is ideal but it's also not ideal to have a one off test following it's own pattern.

.changes/unreleased/Under the Hood-20230203-191815.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Under the Hood
+body: Test conversion for snapshot tests
+time: 2023-02-03T19:18:15.11652-05:00
```
nssalian, line 3: Don't need a changie for test conversions. Can remove.

tests/functional/adapter/common.py
```diff
@@ -0,0 +1,136 @@
+from typing import Dict, Tuple, List
```
nssalian, line 1: typically something termed `common.py` would be used across the board or at least in multiple places. Do we intend to use this for other tests? 
If we don't, we should move this inside the `snapshot_tests` directory with a name like `table_operations.py` or something similar since it's tied to just this set of tests.

tests/functional/adapter/snapshot_tests/files.py
```diff
@@ -0,0 +1,68 @@
+SEED_CSV = """
```
nssalian, line 1: `fixtures.py` might be a better name here.

nssalian (commented):
Some comments over file naming.
Otherwise, I'd still recommend the changes from @colin-rogers-dbt.

Question: I'm finding it challenging to map the tests outlined in `test_simple_snapshot.py` and `test_snapshot_check_cols.py` over to these. I assume we covered all the checks here because it does seem like there was some consolidation. I'm hoping we didn't miss out on some of them.

tests/functional/adapter/test_snapshot.py
mikealfare: I moved this into a separate file.

tests/functional/adapter/test_snapshot.py
mikealfare: After an internal discussion with the team, we sad we'd move forward with this approach in this scenario and submit a proposal to follow this moving forward.

tests/functional/adapter/common.py
mikealfare: Yes, the intention would be to use these across all of the tests. I would have called it `utils` if we didn't already have a directory called that. That directory is actually testing utilities, not utilities used for testing; so perhaps this directly should renamed `utils_tests` and this can be named `utils.py`.

tests/functional/adapter/snapshot_tests/files.py
mikealfare: This file contains contents that are inserted into fixtures, but aren't themselves fixtures. When looking at the corresponding tests in `dbt-core`, the name `files.py` was used for the same concept.

.changes/unreleased/Under the Hood-20230203-191815.yaml
mikealfare: Will do

tests/functional/adapter/snapshot_tests/files.py
mikealfare: I split `files.py` into `seedds.py` and `snapshots.py` to be consistent with how other adapter repos have handled this. That left me with a very small model that I just put in the test file itself instead of creating a separate file for it.

tests/functional/adapter/snapshot_tests/test_snapshot.py
```diff
@@ -0,0 +1,205 @@
...
+
+class SnapshotBase:
+
+    @pytest.fixture(scope="class")
+    def seeds(self):
+        """
+        This seed file contains all records needed for tests, including records which will be inserted after the
+        initial snapshot. This makes it so that Redshift creates the correct size varchar columns. This table
+        will only need to be loaded once at the class level. It will never be altered, hence requires no further
+        setup or teardown.
```
colin-rogers-dbt, line 26: we do drop these artifacts from the db at the end of the test run right?

tests/functional/adapter/snapshot_tests/seeds.py
```diff
@@ -0,0 +1,33 @@
+SEED_CSV = """
```
colin-rogers-dbt, line 1: nit: we've typically been using `fixtures.py`, maybe nest these files under a fixtures directory or add `_fixtures` as a suffix?

tests/functional/adapter/snapshot_tests/test_snapshot.py
```diff
@@ -0,0 +1,205 @@
...
+    def delete_snapshot_records(self):
+        common.delete_records(self.project, "snapshot")
+
+    def _assert_results(
+            self,
+            ids_with_current_snapshot_records: Iterable,
+            ids_with_closed_out_snapshot_records: Iterable
+    ):
+        """
+        All test cases are checked by considering whether a source record's id is end dated in `snapshot`. Each id
```
colin-rogers-dbt, line 88: nit: not sure what `end dated` means here

colin-rogers-dbt (commented):
A couple nits and questions

tests/functional/adapter/snapshot_tests/files.py
colin-rogers-dbt: missed this convo on my review

tests/functional/adapter/snapshot_tests/test_snapshot.py
mikealfare: Oh, I assumed that the `project` fixture handled that. We don't even specify the schema. So I assumed it was created, yielded, and then cleaned up upon test completion. But I can check.

tests/functional/adapter/snapshot_tests/test_snapshot.py
mikealfare: Sorry, vernacular from a previous company. End dated in this context means the record's valid date is populated with the run time, thereby "closing out" the record or providing and "end date". I can update the comment with dbt-specific terminology.

tests/functional/adapter/snapshot_tests/files.py
mikealfare: Is everyone comfortable with these file names then?

tests/functional/adapter/snapshot_tests/test_snapshot.py
mikealfare: The project is yielded out of the fixture here: https://github.com/dbt-labs/dbt-core/blob/5ddd40885e0502e968fd2633127cbb7a1959a40f/core/dbt/tests/fixtures/project.py#L456. After the yield, the teardown attempts to drop the schema. It points out cases where that command may not work and references this issue to resolve those cases: https://github.com/dbt-labs/dbt-core/issues/5041. I would rather solve for it in general with the teardown in `project()` instead of in this PR. Let me know if you feel differently.

colin-rogers-dbt (approved):

PR #323, opened 2023-02-16
Support additional connection methods

### LINKED ISSUE #332: [CT-2156] [Feature] Add dbt-redshift connection methods - Auth Profile and IdP Credentials(Azure and Okta Identity Providers)

### Describe the feature

Provide additional authentication methods to connect to dbt-redshift. The new code changes will allow users to connect to dbt-redshift using authentication profiles and Identity Provider (IdP) Plugins for Azure and Okta. Users will need to provide required parameters specified in [this Redshift documentation](https://docs.aws.amazon.com/redshift/latest/mgmt/odbc20-authentication-ssl.html) when configuring profiles.yml in order to authenticate using these new methods.

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

jiezhen-chen:
resolves #332

### Checklist


dbt/adapters/redshift/connections.py
```diff
@@ -59,7 +63,19 @@ class RedshiftCredentials(Credentials):
     connect_timeout: int = 30
     role: Optional[str] = None
     sslmode: Optional[str] = None
+    application_name: Optional[str] = "dbt"
```
colin-rogers-dbt, line 66: this should be a constant and not settable in the connection configuration

dbt/adapters/redshift/connections.py
```diff
@@ -39,13 +40,16 @@ def json_schema(self):
...
+    AUTH_PROFILE = "auth_profile"
+    IDP = "IdP"
 
 
 @dataclass
 class RedshiftCredentials(Credentials):
-    host: str
-    user: str
     port: Port
+    region: Optional[str] = None
```
colin-rogers-dbt, line 50: Can we infer from the host value the region here like `host.split(".")[2]`?

Fleid:
Hey @jiezhen-chen could you please create an issue getting a bit of context and documenting what's being added?
Mainly for the public record.

Thanks :)

dbt/adapters/redshift/connections.py
jiezhen-chen: since host isn't required anymore(auth_profile doesn't require host) setting region = host.split(".")[2] will fail. What I've done is setting region to host.split(".")[2] in the get_connec_method() call if the connection method is iam or database

dbt/adapters/redshift/connections.py
```diff
@@ -84,14 +108,18 @@ def __init__(self, credentials):
...
             "port": self.credentials.port if self.credentials.port else 5439,
             "auto_create": self.credentials.autocreate,
             "db_groups": self.credentials.db_groups,
-            "region": self.credentials.host.split(".")[2],
             "timeout": self.credentials.connect_timeout,
+            "application_name": "dbt",
         }
+        if method == RedshiftConnectionMethod.IAM or method == RedshiftConnectionMethod.DATABASE:
+            kwargs["host"] = self.credentials.host
+            kwargs["region"] = self.credentials.host.split(".")[2]
```
colin-rogers-dbt, line 122: if the user passes a region that's different from the region in the host should we raise an exception?

dbt/adapters/redshift/connections.py
```diff
@@ -107,14 +135,115 @@ def get_connect_method(self):
...
+                    db_user=self.credentials.user,
+                    auth_profile=self.credentials.auth_profile,
+                    **kwargs,
                 )
                 if self.credentials.role:
                     c.cursor().execute("set role {}".format(self.credentials.role))
                 return c
 
             return connect
+        elif method == RedshiftConnectionMethod.IDP:
```
colin-rogers-dbt, line 174: have we done any local testing of these new connection methods? Ideally we would be able to cover these with integration tests but that may be unrealistic

dbt/adapters/redshift/connections.py
```diff
@@ -107,14 +135,115 @@ def get_connect_method(self):
...
+                        password=self.credentials.password,
+                        idp_tenant=self.credentials.azure_idp_tenant,
+                        client_id=self.credentials.azure_client_id,
+                        client_secret=self.credentials.azure_client_secret,
+                        preferred_role=self.credentials.azure_preferred_role,
+                    )
+                    return c
+
+                return connect
+            elif self.credentials.credentials_provider == "OktaCredentialsProvider":
```
colin-rogers-dbt, line 220: Same comment as above about testing for IDP

dbt/adapters/redshift/connections.py
```diff
@@ -107,14 +135,115 @@ def get_connect_method(self):
...
+                    )
+
+                def connect():
+                    logger.debug("Connecting to redshift with Okta Credentials Provider...")
+                    c = redshift_connector.connect(
+                        iam=True,
+                        region=self.credentials.region,
+                        database=self.credentials.database,
+                        cluster_identifier=self.credentials.cluster_id,
+                        credentials_provider="OktaCredentialsProvider",
```
colin-rogers-dbt, line 237: nit: does this need to be hardcoded? Looks like we already check that self.credentials.credentials_provider == "OktaCredentialsProvider" above

colin-rogers-dbt (commented):
A few small comments, need to think about how we can add some functional testing on the new connection methods to ensure future changes don't break this

dbt/adapters/redshift/connections.py
colin-rogers-dbt: makes sense!

dbt/adapters/redshift/connections.py
jiezhen-chen: Good catch, we could also default to using region from host, in the scenario that host is provided. So if provided region and region in host are different we use the region from host. But we could also throw an error. What are your thoughts?

dbt/adapters/redshift/connections.py
jiezhen-chen: Yes, I've done local testing for these new connection methods. What I've done for local testing - Got some credentials from someone who manages redshift_connector and edited the profiles.yml to connect to Redshift.

jiezhen-chen: Not sure how feasible integration tests are - since we got these credentials from another team and we don't own these credentials. Ideally we should spend some time to create these credentials ourselves so that the integration tests don't depend on another team - but that may be out of scope for right now.

dbt/adapters/redshift/connections.py
jiezhen-chen: Good call. Might also be a good idea to allow more wiggle room for the customer to use these credentials. I think we can hardcode credentials_provider="OktaCredentialsProvider" in the connection call, and allow for multiple variations of Okta from the customer(both "okta" and "oktacredentialprovider" should be accepted with case insensitivity), instead of requiring customers to input excatly "OktaCredentialsProvider"

jaklan:
Hello guys, any chance to push it forward? We really miss the support for the [generic SAML Browser IdP plugin](https://github.com/aws/amazon-redshift-python-driver/blob/master/tutorials/001%20-%20Connecting%20to%20Amazon%20Redshift.ipynb), but the MR seems to be stale for months 😕

misteliy:
Any update here? Especially the generic IdP is very much needed in our case - would be great to see this moving forward.

jaklan:
Bump

PR #39, opened 2021-11-08
Fix tests failing due to dbt-core changes

jtcohen6:
Does not resolve #36, so `test_concurrent_transaction` will continue to fail. Silence the tests for now

PR #325, opened 2023-02-17
[CT-2099]: Delete concurrent_transaction test and clean up integration tests

nssalian:
resolves #

### Description

Deleting concurrent_transaction test and clean up integration tests 

### Checklist


tests/functional/adapter/concurrent_transaction/test_concurrent_transaction.py
```diff
@@ -0,0 +1,128 @@
...
+    def __init__(self):
+        self.query_state = None
+
+    def reset(self):
+        self.query_state = {
+            'view_model': 'wait',
+            'model_1': 'wait',
+        }
+
+    @pytest.fixture(scope="class")
```
mikealfare, line 19: I don't think this will run unless you have `autouse=True` in `fixture()`.

tests/functional/adapter/concurrent_transaction/test_concurrent_transaction.py
```diff
@@ -0,0 +1,128 @@
...
+        return "concurrent_transaction"
+
+    @pytest.fixture(scope="class")
+    def macros(self):
+        return {
+            "udfs.sql": create_udfs_sql
+        }
+
+    def run_select_and_check(self, rel, sql, project):
+        connection_name = '__test_{}'.format(id(threading.current_thread()))
```
mikealfare, line 34: This can be an f-string.

tests/functional/adapter/concurrent_transaction/test_concurrent_transaction.py
```diff
@@ -0,0 +1,128 @@
...
+                client = project.adapter.connections \
+                    .get_thread_connection().handle
+                res = project.run_sql(sql, 'one', client)
+
+                if res[0]:
+                    self.query_state[rel] = 'good'
+                else:
+                    self.query_state[rel] = 'bad'
+
+        except Exception as e:
```
mikealfare, line 46: Generally catching all Exceptions is bad practice. Can we catch the one that we expect? Better than that, if we expect an error, can we assert that we get that error in that scenario? Otherwise there should be no exception catch in the test case. Also, the if block is not needed. The two conditions result in the same code.

tests/functional/adapter/concurrent_transaction/test_concurrent_transaction.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def async_select(self, rel, project, sleep=10):
+        # Run the select statement in a thread. When the query returns, the global
+        # query_state will be updated with a state of good/bad/error, and the associated
+        # error will be reported if one was raised.
+
+        schema = project.test_schema
+        query = '''
+        -- async_select: {rel}
+        select {schema}.f_sleep({sleep}) from {schema}.{rel}
```
mikealfare, line 60: This should be an f-string.

tests/functional/adapter/concurrent_transaction/test_concurrent_transaction.py
```diff
@@ -0,0 +1,128 @@
...
+            'view_model': 'wait',
+            'model_1': 'wait',
+        }
+
+    @pytest.fixture(scope="class")
+    def setUp(self, project):
+        self.reset()
+
+    @pytest.fixture(scope="class")
+    def schema(self):
```
mikealfare, line 24: Do we need to specify the schema? Or can we just use the random unique one that the framework creates for each run?

tests/functional/adapter/concurrent_transaction/test_concurrent_transaction.py
```diff
@@ -0,0 +1,128 @@
...
+from dbt.tests.util import run_dbt
+from dbt.tests.adapter.simple_seed.test_seed import SeedConfigBase
+from tests.functional.adapter.concurrent_transaction.fixtures import *
+
+
+class BaseConcurrentTransaction(SeedConfigBase):
+    def __init__(self):
+        self.query_state = None
+
+    def reset(self):
```
mikealfare, line 13: Does this get called in `SeedConfigBase` somewhere? Otherwise I think it's only called once, in `def setUp()`, which means it can just be moved there in place of the method call.

tests/functional/adapter/concurrent_transaction/test_concurrent_transaction.py
```diff
@@ -0,0 +1,128 @@
+import pytest
+import threading
+
+from dbt.tests.util import run_dbt
+from dbt.tests.adapter.simple_seed.test_seed import SeedConfigBase
+from tests.functional.adapter.concurrent_transaction.fixtures import *
+
+
+class BaseConcurrentTransaction(SeedConfigBase):
+    def __init__(self):
```
mikealfare, line 10: If `query_state` is always set in `def setUp()`, I don't think this is needed.

tests/functional/adapter/concurrent_transaction/fixtures.py
```diff
@@ -0,0 +1,60 @@
...
+-- incremental model
+select 1 as id
+
+{% if is_incremental() %}
+    where TRUE
+{% endif %}
+""".lstrip()
+
+incremental_view_sql = """
+select * from {{ ref('model_1') }}
```
mikealfare, line 31: The three views are the same model. Do we need to have all three? I think the idea is that this model can be built regardless of the materialization of the dependent model. So we just need to define it once and then swap out the dependent model for the three tests. That would also make the block comments unnecessary.

tests/functional/adapter/concurrent_transaction/test_concurrent_transaction.py
nssalian: I thought so, but this test failed because it wasn't initialized: https://github.com/dbt-labs/dbt-redshift/actions/runs/4199862373/jobs/7285257800

tests/functional/adapter/concurrent_transaction/test_concurrent_transaction.py
nssalian: SeedConfigBase is pretty basic - just a `project_config` definition.

tests/functional/adapter/concurrent_transaction/test_concurrent_transaction.py
nssalian: Let me test if this isn't needed.

tests/functional/adapter/concurrent_transaction/test_concurrent_transaction.py
mikealfare: Is that possibly related to the `autouse=True` piece? I was seeing similar behavior until I turned that on.

nssalian: I'll try that and post back.

nssalian: DIdn't help.  AttributeError: 'TestTableConcurrentTransaction' object has no attribute 'query_state'
in the latest test run failure. I don't see `autouse` used in other tests and it seems to have worked fine.

tests/functional/adapter/concurrent_transaction/test_concurrent_transaction.py
mikealfare: Yeah, then I wouldn't bother defining this outside of `def setUp()`. `setUp` already indicates that you're resetting.

nssalian: I doubt that gets much of a benefit. `reset` is more explicit. It's more readable, IMO. I can add a docstring?
```
    def setUp(self, project):
        # Resetting the query_state
        self.query_state = {
            'view_model': 'wait',
            'model_1': 'wait',
        }
```
but not so different from what we have.

tests/functional/adapter/concurrent_transaction/test_concurrent_transaction.py
mikealfare: When I try to run this locally:

```
(venv) mikealfare@dbtlabs dbt-redshift % python -m pytest tests/functional/adapter/concurrent_transaction/test_concurrent_transaction.py -v
```

I get this:

```
collected 0 items

=== warnings summary ====
tests/functional/adapter/concurrent_transaction/test_concurrent_transaction.py:84
  /Users/mikealfare/Source/dbt-redshift/tests/functional/adapter/concurrent_transaction/test_concurrent_transaction.py:84: PytestCollectionWarning: cannot collect test class 'TestTableConcurrentTransaction' because it has a __init__ constructor (from: tests/functional/adapter/concurrent_transaction/test_concurrent_transaction.py)
```

If you need that attribute to exist, can you maybe make it a class attribute then? Otherwise I don't think these tests will run, and then they'll look like they pass as a a result.

nssalian:
Update: This test is still failing. The issue was fixed back in 2018: https://github.com/dbt-labs/dbt-core/pull/825
This is the bug issue it resolved: https://github.com/dbt-labs/dbt-core/issues/653.
Worth getting the test in but it is failing on the udfs

tests/functional/adapter/concurrent_transaction/test_concurrent_transaction.py
```diff
@@ -0,0 +1,120 @@
+import time
+
+import pytest
+import threading
+
+from dbt.tests.util import run_dbt
+from dbt.tests.adapter.simple_seed.test_seed import SeedConfigBase
```
McKnight-42, line 7: potentially unneeded import

tests/functional/adapter/concurrent_transaction/test_concurrent_transaction.py
```diff
@@ -0,0 +1,120 @@
...
+from tests.functional.adapter.concurrent_transaction.fixtures import *
+
+
+class BaseConcurrentTransaction:
+    @pytest.fixture(scope="function", autouse=True)
+    def setUp(self, project):
```
McKnight-42, line 13: do we want this to be the setup or should re rename back to `reset` like the integration test

tests/integration/concurrent_transaction_test/test_concurrent_transaction.py
```diff
@@ -1,136 +0,0 @@
-from tests.integration.base import DBTIntegrationTest, use_profile
-import threading
-from dbt.adapters.factory import FACTORY
-
-
-def get_adapter_standalone(config):
-    plugin = FACTORY.plugins[config.credentials.type]
-    cls = plugin.adapter
-    return cls(config)
```
McKnight-42, line 9: does the new functional test need this?

nssalian: I tried adding it but it didn't make a difference. The error was in the underlying udf so I wonder if it might be a race condition.

tests/integration/concurrent_transaction_test/test_concurrent_transaction.py
```diff
@@ -1,136 +0,0 @@
...
-    def reset(self):
-        self.query_state = {
-            'view_model': 'wait',
-            'model_1': 'wait',
-        }
-
-    def setUp(self):
-        super().setUp()
-        self._secret_adapter = get_adapter_standalone(self.config)
-        self.reset()
```
McKnight-42, line 23: just a ref for comment above about renaming of setUp to reset

tests/functional/adapter/concurrent_transaction/test_concurrent_transaction.py
nssalian: @mikealfare suggested changing this earlier. Went back and forth on it. I don't have a preference. I'll get it to reset() for now to be on par with the previous integration.

McKnight-42: i'd have to try to find the context but I think setUp has some special context for triggering namewise as well so might be kicking stuff off when we don't want to

McKnight-42:
Do we any insight into if the issue that caused us to silence these tests is fixed or if we can even get these up and working?

nssalian:
Context: I checked with @jtcohen6 and we agreed to delete these tests since they were commented out since 2021 via [PR](https://github.com/dbt-labs/dbt-redshift/pull/39). 
This removes the integration tests completely.

McKnight-42 (approved):

PR #329, opened 2023-02-21
Test updating dev-requirements.txt / setup.py on workflow dispatch

colin-rogers-dbt:
Install core dependencies for integration testing from dbt-core branch.

### Checklist


colin-rogers-dbt:
Manually kicked off test run: https://github.com/dbt-labs/dbt-redshift/actions/runs/4235851680/jobs/7359978550

.github/workflows/integration.yml
```diff
@@ -34,6 +34,11 @@ on:
   # manual trigger
   workflow_dispatch:
   # run this once per night to ensure no regressions from latest dbt-core changes
+    inputs:
+      dbt-core-branch:
+        description: "branch of dbt-core to use in dev-requirements.txt"
+        required: false
+        type: string
   schedule:
```
emmyoop, line 42:
```suggestion
  workflow_dispatch:
    inputs:
      dbt-core-branch:
        description: "branch of dbt-core to use in dev-requirements.txt"
        required: false
        type: string
  # run this once per night to ensure no regressions from latest dbt-core changes
  schedule:
```

emmyoop (changes requested):
Won't this also require always having a branch in `dev-requirements.txt`, even if it's just `main`?

colin-rogers-dbt:
> Won't this also require always having a branch in `dev-requirements.txt`, even if it's just `main`?

Actually putting a branch in dev-requirements would break this since I just prepend branch name i.e. you would end up with:
`git+https://github.com/dbt-labs/dbt-core.git@main@main#egg=dbt-core&subdirectory=core`

emmyoop:
But then we can never run this for workflows that have a base of one of the latest branches.  Are we okay with that?  It's not common, but not unheard of either.

emmyoop (approved):
Approving knowing this may not work for windows.  Since it can only be invoked on purpose nothing current will break.  You can't actually test this until it's in main anyways,

colin-rogers-dbt:
> But then we can never run this for workflows that have a base of one of the latest branches. Are we okay with that? It's not common, but not unheard of either.

Good call out, update the sed syntax to handle this! 

> Approving knowing this may not work for windows. 

Looks like it does run on Windows! 

> You can't actually test this until it's in main anyways.
Since this is an update to an existing workflow I am able to test it off a branch with:
`gh workflow run integration.yml -f dbt-core-branch=cl/dbt-retry --ref update_dev_req`
See: https://github.com/dbt-labs/dbt-redshift/actions/runs/4237292878/jobs/7363103011

PR #345, opened 2023-02-28
work on adapter cut-release and nightly-release update

McKnight-42:
resolves #https://github.com/dbt-labs/actions/issues/76

### Checklist


McKnight-42:
workflow chart https://github.com/dbt-labs/dbt-redshift/actions/runs/4325439845
changelog: branch pr: https://github.com/dbt-labs/dbt-redshift/pull/354
currently existing test branch: https://github.com/dbt-labs/dbt-redshift/tree/1.5.test

.github/scripts/update_dependencies.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash -e
+set -e
+
```
mikealfare, line 3: I left a comment in #dbt-labs/actions/83 that probably belongs here. Once this hits `main`, we would want to point to `main` on `dbt-core`, but from `setup.py`, right? And then the requirement in `dev-requirements.txt` should just be removed, since it will come in via `pip install -e .`. And that would ensure that we're also not trying to install two versions of `dbt-core`. Likewise, there should be no requirement for `dbt-core` in `dev-requirements.txt` on any `1.x` version, and `setup.py` would reflect a requirement of `dbt-core~=1.x`. Is that correct?

emmyoop: How would you point to branches while developing in this model?

.github/workflows/cut-release-branch.yml
```diff
@@ -32,11 +32,11 @@ permissions:
 jobs:
   cut_branch:
     name: "Cut branch and clean up main for dbt-redshift"
-    uses: dbt-labs/actions/.github/workflows/cut-release-branch.yml@main
+    uses: dbt-labs/actions/.github/workflows/cut-release-branch.yml@mcknight/update_dependencies
```
emmyoop, line 35: This needs to be updated before merging.

emmyoop (approved):
Approving - just need to update the workflow to point back to main when your'e done testing..

.github/scripts/update_dependencies.sh
mikealfare: I can only think of one situation where we would need a different version of `dbt-core` than `main`; that's if we're developing something that requires updates to both `dbt-core` and `dbt-<adapter>`. If there's another scenario, please mention it. I'll assume not for the rest of this comment though.

I would think we'd want to minimize how often we update both `dbt.core` and `dbt-<adapter>` in a way where they are dependent on each other. If we're saying we need to release to both simultaneously, i.e. we're going to break `dbt-<adapter>` with an update to `dbt-core`, then we're also saying that we're going to break all of the third-party adapter maintainers.

Updates to `dbt-core` that _require_ updates to `dbt-<adapter>` reflect a change in interface; they would affect all adapters, and should be thought about in that context. Whereas updates to `dbt-<adapter>` would only affect that adapter, hence have different concerns. If we do need to make updates to `dbt-core` for `dbt-<adapter>`, I would argue that they should be tested and merged into `main` on `dbt-core` first.

I understand that's more idealistic than we can expect for every body of work. So in the (rare) case that I would need to have parallel branches in `dbt-core` and `dbt-adapter`, I would manually update either `dev-requirements.txt` or `setup.py`. And I'd probably choose the latter since that's where the dependency ultimately lives. If we need a programmatic way of switching branches (e.g. we do this often for an adapter for a reason I'm missing), then I would update the function that determines the `dbt-core` version in `setup.py` to accept an override.

tdlr;
I effectively would not point to branches while developing unless absolutely necessary. In the event that I need to do that, I would manually add it to `setup.py` or update the `dbt-core` version function in `setup.py` with an override.

emmyoop: Two more cases come to mind.

1.  The only other case I can think of is for our `*.latest` branches.  Because of our release/branching strategy, once we cut the `*.latest` branch, it's the truth for that version and what's on `main` is no longer really relevant.  If we need a change on that`*.latest` branch, we have to deliberately backport it.  For this scenario, it's really safer to be testing the adapters against their equivalent `*.latest` branch in `dbt-core` so that when we do release patches we're not surprised by anything breaking.  

2.  There are also times where we are making changes that are adding functionality and we want to ensure it works on adapters before merging everything into `dbt-core` `main`.  If we were not able to test the branch of `dbt-core` against new functionality, it would cause slowdown in development and risk of committing broken code into `dbt-core`.

I 100% agree that changes to core that break adapters should be considered interface changes and we should really consider if they should be made and how.

As long as there is a way to override, all the cases we came up with continue to work.

mikealfare: > 1. The only other case I can think of is for our *.latest branches. Because of our release/branching strategy, once we cut the *.latest branch, it's the truth for that version and what's on main is no longer really relevant. If we need a change on that*.latest branch, we have to deliberately backport it. For this scenario, it's really safer to be testing the adapters against their equivalent *.latest branch in dbt-core so that when we do release patches we're not surprised by anything breaking.

Talking "out loud" here. So in this case, the version of `dbt-core` in `dev-requirements.txt` is ahead of the version of `dbt-core` in `setup.py` because `dev-requirements.txt` is looking at GitHub whereas `setup.py` is looking at PyPI. So we wouldn't be able to really test this without intervention until the relevant version of `dbt-core` is released to PyPI, which may take time. Is all of that correct? And if so, is there any way to guarantee that we're testing `dbt-<adapter>` against the relevant production (PyPI) version of `dbt-core` before releasing?

> 2. There are also times where we are making changes that are adding functionality and we want to ensure it works on adapters before merging everything into dbt-core main. If we were not able to test the branch of dbt-core against new functionality, it would cause slowdown in development and risk of committing broken code into dbt-core.

Agreed, new functionality without changing existing functionality would require something like this. In that case, I would think of it as two problems:

1. `dbt-core` needs new (abstract) functionality, and needs to test that it works in an implementation 
2. `dbt-<adapter>` would like to implement that new functionality, and needs it in `dbt-core` to continue

The second item is potentially a way to test the first item, but it's not the only way. This is far beyond the scope of this PR, but at some point we need to mock out all of the adapter base classes so that we can test an implementation in `dbt-core` without requiring updates in an adapter repo. This heavy coupling is problematic.

Getting back to this particular PR, I suppose keeping a permanent link in `dev-requirements.txt` is not the worst idea. But in the future, we should look at decoupling this.

mikealfare (approved):

Issue #204, opened 2022-10-13
[CT-1340] [Feature] Support Python models (dbt-py) on Redshift/AWS


ChenyuLInx:
### Is this your first time submitting a feature request?

- [X] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)
- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward extension of existing dbt-redshift functionality, rather than a Big Idea better suited to a discussion

## Background:

There's a [Spark redshift connector](https://docs.aws.amazon.com/emr/latest/ReleaseGuide/emr-spark-redshift.html). This would allow user to run python transformation code on EMR cluster that load data from Redshift, and write transformed data back to Redshift. The whole process is very similar to using Dataproc to run python models on GCP/BigQuery.

## Items needed for implementation:
 
- If there's additional profile information needed for EMR cluster, we can add it as optional attributes at [Credentials](https://github.com/dbt-labs/dbt-redshift/blob/main/dbt/adapters/redshift/connections.py#L41)(existing example for [bigquery](https://github.com/dbt-labs/dbt-bigquery/blob/main/dbt/adapters/bigquery/connections.py#L123)).
- We need one macro to generate the final code to run on EMR cluster, Previous example for dbt-bigquery [here](https://github.com/dbt-labs/dbt-bigquery/blob/main/dbt/include/bigquery/macros/materializations/table.sql#L51)
- Now that we have the profile info and the macro to generate final code, we need submission classes to submit python code to the cluster. [Existing submission code for dbt-bigquery](https://github.com/dbt-labs/dbt-bigquery/blob/main/dbt/adapters/bigquery/python_submissions.py), Function to define in impl.py([link1](https://github.com/dbt-labs/dbt-bigquery/blob/main/dbt/adapters/bigquery/impl.py#L851), [link2](https://github.com/dbt-labs/dbt-bigquery/blob/main/dbt/adapters/bigquery/impl.py#L855)). And how those [classes are being used by dbt-core](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/adapters/base/impl.py#L1223)(This doesn't need to be changed.)

colin-rogers-dbt:
I wonder if the right thing here to do is just pick on to implement with a long term of supporting both, can we leverage the existing spark/glue adapters?

lostmygithubaccount:
[...]

PR #379, opened 2023-03-21
ADAP-381: adds query tagging support and materializations to dbt-redshift

### LINKED ISSUE #376: Enable tagging of query_group in configs

### Describe the feature

Much akin to dbt-snowflake's implementation of [query tagging](https://docs.getdbt.com/reference/resource-configs/snowflake-configs#query-tags), I think it'd be great to allow users to set a [query_group](https://docs.aws.amazon.com/redshift/latest/dg/r_query_group.html) for their models. I'd imagine the implementation of this would be quite straightforward as the functionality is quite similar to that of Snowflake's, but hoping to get input there.

Not aware that there are any alternatives.

### Who will this benefit?

Anybody who uses the Redshift query logs to perform analysis on their queries.

Certainly! Feels like it could be a good first issue.

### Anything else?

_No response_

trouze:
resolves #376 

### Description

This PR will add the ability to tag queries ran against Redshift with a string field added by a user in yaml or sql configs, much like the dbt-snowflake adapter supports.

Tradeoffs:
- To support this, I've also added materialization files to the adapter since the base dbt-core materializations don't support query tagging (if it is desired to go the other direction and have dbt-core support this please let me know).
- The decision was made to keep the naming of `query_tag` consistent with that of dbt-snowflake, despite AWS naming this feature `query_group`. We can certainly go back on this convention, but obviously would recommend keeping the macros consistent with that of dbt-snowflake to support multiple dispatch.

Pertinent links to this functionality and dbt-snowflake's implementation of this feature can be found in the issue discussion.

### Checklist


VersusFacit:
Heya, I see you're adding several local macros in order to get this done. That's potentially all right, but that also means the reviews going to take longer because I need to evaluate some design considerations. 

The actual feature itself seems simple enough and as you pointed out, has precedence in snowflake.

boxysean:
Hi @trouze @VersusFacit, I reviewed this for about 15 minutes. Agreed it's a large PR primarily because of the requirement of bringing in `redshift__*` materializations into the dbt-redshift project. As @trouze pointed out, the decision we must make is whether to take this route, or instead bring int the query-tagging concept into dbt-core and utilize it much easier in dbt-redshift (i.e., this PR changed line size would drop by ~500 lines).

Check out this thread going on https://github.com/dbt-labs/dbt-bigquery/issues/685 -- @dbeatty10 do you think this PR from @trouze is sufficiently similar?

trouze:
@boxysean thank you for having a look at this! I think it makes sense to have dbt-core utilize the query tagging concept in the base materializations there, and then have each adapter implement macros specific to how their database implements the query tagging functionality. Despite @dbeatty10 mention that `redshift__*` materializations will be introduced with #204, I think this approach is preferable. I'd be happy to open an issue over on dbt-core for this and cut down this PR to match if that's the direction we'd like to go, but will wait for further guidance here.

ddors3y:
This feature has came back up as something we are really interested in. Looks like it hasn't got much activity lately so wanted to check and see if there was another place this got moved or if there was alternative.

trouze:
@ddors3y I believe we were all in agreement with moving this to `dbt-core`, I opened a [draft illustration](https://github.com/dbt-labs/dbt-core/pull/7770) of what it might look like there but just never found the time to bring it to completion. I may get back around to it coming up here though.

PR #386, opened 2023-03-28
ADAP-2: Materialized Views

mikealfare:
resolves https://github.com/dbt-labs/dbt-core/issues/6911

### Description

Implement materialized views as a materialization option for Redshift.

Workflow:
- Create a new materialized view
- Replace an existing materialized view on full refresh
- Replace an existing object if it is not a materialized view, regardless of full refresh
- Refresh existing materialized view with no changes
- Accept and process changes to a materialized view:
  - `sort`: full refresh
  - `dist`: full refresh
  - `auto_refresh`: alter materialized view in place

Configuration:
- `on_configuration_change`: determines the action to take when changes are detected:
  - `apply`: full refresh on sortkey/distkey changes, alter on auto_refresh changes
  - `continue`: soft fail, skip the processing of this model, but process downstream
  - `fail`: fail entire run
- `auto_refresh`: turns on/off auto refresh of the materialized view

### Checklist


dbt/adapters/redshift/impl.py
```diff
@@ -27,13 +27,13 @@ class RedshiftConfig(AdapterConfig):
...
     bind: Optional[bool] = None
     backup: Optional[bool] = True
+    autorefresh: Optional[bool] = False
 
 
 class RedshiftAdapter(SQLAdapter):
     Relation = RedshiftRelation
     ConnectionManager = RedshiftConnectionManager
     connections: RedshiftConnectionManager
-    Column = RedshiftColumn  # type: ignore
```
mikealfare, line 36: We did this because we initially inherited from PostgresAdapter, however we wanted the default Column from `dbt-core`. So we redefined it in `dbt-redshift` and overwrote `PostgresAdapter.Column` with (effectively) `SQLAdapter.Column`. Now that we're back to inheriting from `SQLAdapter`, we no longer need this, nor the associated file/class.

dbt/adapters/redshift/relation.py
```diff
@@ -1,13 +1,92 @@
...
-    # Override the method in the Postgres Relation because Redshift allows
-    # longer names: "Be between 1 and 127 bytes in length, not including
-    # quotation marks for delimited identifiers."
-    #
-    # see: https://docs.aws.amazon.com/redshift/latest/dg/r_names.html
+class RedshiftRelation(BaseRelation):
+    include_policy = RedshiftIncludePolicy  # type: ignore
+    quote_policy = RedshiftQuotePolicy  # type: ignore
+
+    def __post_init__(self):
```
mikealfare, line 29: This is the only remaining method in `PostgresRelation` that isn't in `BaseRelation`. Re-defining it and inheriting from `BaseRelation` is less error prone going forward as we now have different definitions for similar methods with materialized views. This also  contributes to a longer term goal of removing `dbt-postgres` as a dependency for `dbt-redshift`.

tox.ini
```diff
@@ -21,7 +21,8 @@ passenv =
     REDSHIFT_TEST_*
     PYTEST_ADDOPTS
 commands =
-  redshift: {envpython} -m pytest {posargs} tests/functional
+  {envpython} -m pytest {posargs} tests/functional -k "not tests/functional/adapter/utils"
```
colin-rogers-dbt, line 24: just curious, why are we separating these out?

dbt/adapters/redshift/relation_configs/base.py
```diff
@@ -0,0 +1,70 @@
...
+from dbt.adapters.redshift.relation_configs.policies import (
+    RedshiftIncludePolicy,
+    RedshiftQuotePolicy,
+)
+
+
+@dataclass(frozen=True, eq=True, unsafe_hash=True)
+class RedshiftRelationConfigBase(RelationConfigBase):
+    """
+    This base class implements a few boilerplate methods and provides some light structure for Snowflake relations.
```
colin-rogers-dbt, line 22: nit: `Snowflake` copy pasta

dbt/adapters/redshift/relation_configs/dist.py
```diff
@@ -0,0 +1,164 @@
...
+
+        elif diststyle == RedshiftDistStyle.key:
+            open_paren = len("KEY(")
+            close_paren = -len(")")
+            distkey = dist[open_paren:close_paren]  # e.g. KEY(column1)
+
+            return {"diststyle": diststyle.lower(), "distkey": distkey}
+
+        else:
+            raise DbtRuntimeError(f"Unexpected dist format: {dist}")
```
colin-rogers-dbt, line 144: Unlikely edge case but do we want to treat an unexpected dist style from redshift as a runtime error? The scenario I'm imagining here being that redshift adds a new diststyle or changes the returned result and dbt jobs will start to fail. 
If we want to treat this as an exception maybe we add a little more detail: `Recieved an unexpected dist format from database: {dist}`?

dbt/include/redshift/macros/materializations/materialized_view.sql
```diff
@@ -0,0 +1,106 @@
...
+            tb.table,
+            tb.diststyle,
+            tb.sortkey1,
+            mv.autorefresh
+        from svv_table_info tb
+        left join stv_mv_info mv
+            on mv.db_name = tb.database
+            and mv.schema = tb.schema
+            and mv.name = tb.table
+        where tb.table ilike '{{ relation.identifier }}'
```
colin-rogers-dbt, line 83: this needs to be `ilike`?

dbt/adapters/redshift/relation_configs/dist.py
```diff
@@ -0,0 +1,164 @@
...
+                    and self.distkey is not None
+                ),
+                validation_error=DbtRuntimeError(
+                    "A `RedshiftDistConfig` that specifies a `distkey` must be of `diststyle` `key`."
+                ),
+            ),
+        }
+
+    @classmethod
+    def from_dict(cls, config_dict) -> "RedshiftDistConfig":
```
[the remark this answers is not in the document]
mikealfare: I'm mirroring frameworks like `mashumaro` since I suspect we might replace this with something like that anyway. Even if we don't, it should evoke those types of feels.

dbt/adapters/redshift/relation_configs/dist.py
mikealfare: Honestly, I think it's more likely that we just won't cover all possible cases that Redshift can produce right out of the gate, rather than there being a truly new diststyle. That being said, I agree the message is vague. I'll update it.

dbt/include/redshift/macros/materializations/materialized_view.sql
mikealfare: I'm accounting for case-insensitivity, but maybe I shouldn't? It's probably worth thinking through that.

tox.ini
mikealfare: We hit the max on the free tier Windows box in CI like we did in `dbt-core` a few sprints ago. So I took the same approach of breaking out the largest test module count-wise. I don't like it, but it works for now, until we figure out how we want to solve for this in general.

dbt/adapters/redshift/relation_configs/base.py
mikealfare: I'm keeping it, lol

dbt/include/redshift/macros/materializations/materialized_view.sql
mikealfare: For now I'm going to keep it since it aligns with `list_relations_without_caching`. But I do think it's worth rethinking it, just at a larger scope than MVs.

dbt/adapters/redshift/relation_configs/dist.py
```diff
@@ -0,0 +1,164 @@
...
+        Args:
+            model_node: the description of the distkey and diststyle from the user in this format:
+
+                {
+                    "dist": any("auto", "even", "all") or "<column_name>"
+                }
+
+        Returns: a standard dictionary describing this `RedshiftDistConfig` instance
+        """
+        dist = model_node.config.extra.get("dist", "")
```
colin-rogers-dbt, line 90:
```suggestion
        dist = model_node.config.extra.get("dist", "").lower()
```
just a nit but would mean you don't need to call `.lower` a couple times later

dbt/adapters/redshift/relation_configs/dist.py
```diff
@@ -0,0 +1,164 @@
...
+        Args:
+            relation_results_entry: the description of the distkey and diststyle from the database in this format:
+
+                agate.Row({
+                    "diststyle": "<diststyle/distkey>",  # e.g. EVEN | KEY(column1) | AUTO(ALL) | AUTO(KEY(id))
+                })
+
+        Returns: a standard dictionary describing this `RedshiftDistConfig` instance
+        """
+        dist: str = relation_results_entry.get("diststyle", "")
```
colin-rogers-dbt, line 119: same suggestion as above

dbt/include/redshift/macros/materializations/materialized_view.sql
```diff
@@ -0,0 +1,106 @@
...
+
+
+{% macro redshift__refresh_materialized_view(relation) -%}
+    refresh materialized view {{ relation }}
+{% endmacro %}
+
+
+{% macro redshift__describe_materialized_view(relation) %}
+    {#-
+        These need to be separate queries because redshift will not let you run queries
```
colin-rogers-dbt, line 66: I'll take seeming random db limitations for 1000

colin-rogers-dbt (approved):
one small suggestion but otherwise LGTM

dbt/include/redshift/macros/materializations/materialized_view.sql
mikealfare: This made my day

dbt/adapters/redshift/relation_configs/dist.py
mikealfare: We merge `diststyle` and `distkey` into one user config of `dist`. In other words, this could be the column which is theoretically possibly case-sensitive. I didn't want to use `.lower()` until I knew it was the `diststyle`.

Less concretely, I also happen to like the separation between "go get the pieces I need" and "do these calculations/formats on the pieces".

dbt/adapters/redshift/relation_configs/dist.py
mikealfare: In this case I forgot to add `.lower()` to the `if` block below, which will fail for keys since Redshift capitalizes it.

I should add it to lines 122 and 124 and remove it afterwards, which aligns with your suggestion of doing it the first place it's calculated.

dbt/adapters/redshift/relation_configs/dist.py
mikealfare: I compromised by defining it in between. I found myself doing that for the other parser and I liked the consistency. I also removed the error check since we're not, nor can we, do that for the node config version. Because it could be the key, the valid value set is only known at the materialized view level (which eventually could also be table, etc).

setup.py
```diff
@@ -86,6 +86,8 @@ def _core_version(plugin_version: str = _plugin_version()) -> str:
         f"dbt-postgres~={_core_version()}",
         "boto3~=1.26.26",
         "redshift-connector~=2.0.911",
+        # installed via dbt-core but referenced directly; don't pin to avoid version conflicts with dbt-core
+        "agate",
```
dataders, line 90: can you help me understand why this is needed? for Pylance linting?

PR #387, opened 2023-03-29
ADAP-389: Stub materialized view as a materialization

mikealfare:
resolves https://github.com/dbt-labs/dbt-core/issues/6911

### Description

Stub out the materialized view materialization for Redshift. Under the hood this just creates a view for now. But this can be used to verify all of the plumbing.

While creating the materialization, it was observed that other materialization templates existed outside of the `materialization` directory. These were moved from `adapters.sql` into the `materialization` directory in the appropriate file (e.g. `table.sql`). Some of the macros were also refactored to be smaller bits of code and `dbt-postgres` dependencies were eliminated when found.

### Checklist


mikealfare:
I'm skipping the change log on this; it's too granular. There will be a changelog at the feature level that will be merged into `main`.

mikealfare:
I'm running into an issue while trying to recreate materialized views. They don't appear to show up as expected in `pg_depend`. When a materialized view is created, call it `my_mat_view`, based on a table called `my_table`, Redshift actually creates a view called `my_mat_view` and a table called `tbl_mv__my_mat_view__0`. The entries that show up in `pg_depend` tie `my_mat_view` to `tbl_mv__my_mat_view__0`, which is technically correct. Unfortunately, `my_table` does not show up at all even though that is the actual dependency from the user's perspective. I believe when we recreate the materialized view, we are inadvertently dropping it because the dependency is not properly managed when recreating the table.

PR #398, opened 2023-04-07
fix the json obj in msgs bug for both text and Json logs

### LINKED ISSUE #373: [ADAP-377] [CT-2308] [Bug] 1.5.0-b3 has funky error logging

- [X] I believe this is a new bug in dbt-core

### Current Behavior

While looking at the debug logs for a dbt utils CI job, I noticed that there's some funky things happening in the logs. 

## 1. Entire JSON object being dumped for database errors: 
Observed:
```
01:59:07  Finished running 67 seeds in 0 hours 0 minutes and 23.98 seconds (23.98s).
01:59:07  
01:59:07  Completed with 1 error and 0 warnings:
01:59:07  
01:59:07  Database Error in seed data_events_20180103 (data/sql/data_events_20180103.csv)
01:59:07    {'S': 'ERROR', 'C': '42501', 'M': 'permission denied for schema dbt_utils_integration_tests_redshift_events', 'F': '../src/pg/src/backend/catalog/aclchk.c', 'L': '2946', 'R': 'aclcheck_error'}
01:59:07  
01:59:07  Done. PASS=66 WARN=0 ERROR=1 SKIP=0 TOTAL=67
```

Expected: 
```
01:59:07  Finished running 67 seeds in 0 hours 0 minutes and 23.98 seconds (23.98s).
01:59:07  
01:59:07  Completed with 1 error and 0 warnings:
01:59:07  
01:59:07  Database Error in seed data_events_20180103 (data/sql/data_events_20180103.csv)
01:59:07    permission denied for schema dbt_utils_integration_tests_redshift_events
01:59:07  
01:59:07  Done. PASS=66 WARN=0 ERROR=1 SKIP=0 TOTAL=67
```

## 2. Wrong output for successful model builds on Redshift
Observed:
```
01:59:13  4 of 47 START sql view model dbt_utils_integration_tests_redshift.test_deduplicate  [RUN]
[...]
01:59:14  4 of 47 OK created sql view model dbt_utils_integration_tests_redshift.test_deduplicate  [cursor.rowcount = -1 in 0.98s]
```

Expected: 
I don't remember offhand, but it sure isn't `cursor.rowcount = -1`! Isn't it meant to be something like `SUCCESS 1` for views and `SUCCESS [impacted row count]` for tables?

### Expected Behavior

Noted above

### Steps To Reproduce

Check out this pipeline for the specific logs https://app.circleci.com/pipelines/github/dbt-labs/dbt-utils/1474/workflows/3b5f8db1-2663-4366-8c53-8a2e7cfb2899/jobs/4810

### Relevant log output

_No response_

### Environment

```markdown
- OS: Ubuntu 20.04.5 LTS
- Python: cimg/python:3.9

Core:
  - installed: 1.5.0-b3
  - latest:    1.4.5    - Ahead of latest version!

Plugins:
  - postgres:  1.5.0b3 - Ahead of latest version!
  - snowflake: 1.5.0b2 - Ahead of latest version!
  - redshift:  1.5.0b2 - Ahead of latest version!
  - bigquery:  1.5.0b2 - Ahead of latest version!
```

redshift

### Additional Context

_No response_

VersusFacit:
resolves #373 

Before too long I'm gonna adopt the title 'lorekeeper of the sacred loggers' what with all these proprietary logger objects being added to my knowledge base

#### Bug 1

Tracked down the root of the unwanted Json block in place of a clean message and extract only the message from that object. 

Here's an example of this bug affecting Json logs before my fix:
image

Post fix for structured logging which also exhibited this issue:
image

I made sure that only the message will show up regardless of what DBT_LOG_FORMAT is set to

#### Bug 2

~~This one is a work in progress. This is talked about in the ticket. I have a problem with the cursor object doesn't seem to contain the row count under the same semantics we used to use. Additionally, no other field seems to capture that behavior. I think we have to do a deeper redesign here but I'm trying to avoid that.~~

update:

image
In the screenshot it's a bit crowded, but you can see the manufactured SUCCESS message that this PR adds. 

image

You can see that we've lost the rowcount, but it doesn't show up in successful queried cursor objects anymore. What strange to me, is that the exact same function during a `seed` operation has access to the correct row count (although to be clear, in that event we don't use _this_ `get_relation` function to render the message `seed` displays – there's a lot of layers of interaction here), so redshift does something to suppress that number; instead they always send back a generic `-1` 🤷‍♀️ especially since every database gives a slightly different message back already

Proof this doesn't break `seed`:
image

### Checklist

Also, I'm gonna have to do a manual backport of 1.4


Fleid:
Sacred Logkeeper?

dbt/adapters/redshift/connections.py
```diff
@@ -178,18 +178,24 @@ def cancel(self, connection: Connection):
 
     @classmethod
     def get_response(cls, cursor: redshift_connector.Cursor) -> AdapterResponse:
+        # redshift_connector.Cursor doesn't have a status message attribute but
+        # this function is only used for successful run, so we can just return a dummy
         rows = cursor.rowcount
-        message = f"cursor.rowcount = {rows}"
+        message = "SUCCESS"
```
VersusFacit, line 184: bug 2

dbt/adapters/redshift/connections.py
```diff
@@ -178,18 +178,24 @@ def cancel(self, connection: Connection):
...
         return AdapterResponse(_message=message, rows_affected=rows)
 
     @contextmanager
     def exception_handler(self, sql):
         try:
             yield
         except redshift_connector.DatabaseError as e:
-            logger.debug(f"Redshift error: {str(e)}")
+            try:
```
VersusFacit, line 192: bug 1: this seems like the most expedient way to default to the exception object if there's any trouble accessing an inner message field

mikealfare: Oof, so it's literally get the first error's "M"essage? They didn't happen to wrap that in anything?

VersusFacit:
I'll make sure these get a green before merging, but yeah it should literally just be needing to merge all the changes that have been shoved into Maine.

Issue #217, opened 2022-11-09
[CT-1482] [Feature] Add support for data share objects


randy-rc:
### Is this your first time submitting a feature request?

- [X] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)
- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward extension of existing dbt-redshift functionality, rather than a Big Idea better suited to a discussion

## Issue 1
**Calling dbt_utils.get_column_values on a data share object results in error**
```
{% macro test5() -%}

    {% set details = dbt_utils.get_column_values(table=source('[name]', '[object]'), column='[column]') %}

    {% for detail in details %}
        {{ log('Detail: ' ~ detail, True) }}
    {% endfor %}

{%- endmacro %}
```

```
09:25:34  Encountered an error while running operation: Compilation Error in macro test5 (macros/test.sql)
  In get_column_values(): relation "[database]"."[schema]"."[object]" does not exist and no default value was provided.
  
  > in macro statement (macros/etc/statement.sql)
  > called by macro default__get_column_values (macros/sql/get_column_values.sql)
  > called by macro get_column_values (macros/sql/get_column_values.sql)
  > called by macro test5 (macros/test.sql)
  > called by macro test5 (macros/test.sql)
```

**dbt Cloud System Logs (Details) captures the statement used to check if object of concern exists. Unfortunately data share objects are not stored in both pg_tables and pg_views and hence the "does not exist" error.**
```
...
09:25:33  On macro_test5: /* {"app": "dbt", "dbt_version": "1.2.2", "profile_name": "user", "target_name": "default", "connection_name": "macro_test5"} */
select
      '[database]' as database,
      tablename as name,
      schemaname as schema,
      'table' as type
    from pg_tables
    where schemaname ilike '[schema]'
    union all
    select
      '[database]' as database,
      viewname as name,
      schemaname as schema,
      'view' as type
    from pg_views
    where schemaname ilike '[schema]'
09:25:34  SQL status: SELECT in 0.24 seconds
09:25:34  with database=[database], schema=[schema], relations=[]
09:25:34  Postgres adapter: Error running SQL: macro test5
...
```

## Issue 2
**Calling adapter.get_columns_in_relation on a data share object does not result in error but no result is returned**
```
{% macro test6() -%}
    
[...]

Issue #179, opened 2022-09-02
[CT-1122] [CT-1116] [Bug] `adapter.get_relation` for relations in other databases [RA3]


jaklan:
### Is this a new bug in dbt-core?

- [X] I believe this is a new bug in dbt-core
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

`adapter.get_relation` seems to ignore `database` argument and use the target database anyway.

### Expected Behavior

The right database is used.

### Steps To Reproduce

- I have a model, defined as below, which checks if it already exists in the same schema, but in another table. If it exists - it prints a log message and reads data from there:
  ```
  {{ config({ "materialized":"table" }) }}
  
  {%- set relation = adapter.get_relation(
      database="integrated",
      schema=this.schema,
      identifier=this.identifier) -%}
  
  {% if relation is not none %}

  {{ log(relation.database ~ '.' ~ relation.schema ~ '.' ~ 
         relation.identifier ~ " already exists", info=true) }}

  select * from {{relation}}

  {% else %}

  ... (e.g. creates an empty table or do nothing if exists)

  {% endif %}
  ```
- I have 2 databases: `dev` and `integrated`. 
- In `dev` I have that table already created, in `integrated` - not. 
- The database specified in the active target is `dev`.
- Now, when running `dbt run -m "<model_identifier>"`, I would expect it to do nothing in `dev` database. But, instead of that I get:

   ```
   02:41:51  1 of 1 START table model some_schema.some_model ................... [RUN]
   02:41:52  integrated.some_schema.some_model already exists
   02:41:52  1 of 1 ERROR creating table model some_schema.some_model .......... [ERROR in 0.80s]

   02:41:55  Completed with 1 error and 0 warnings:
   02:41:55  
   02:41:55  Database Error in model <model_identifier>
   02:41:55    Schema some_schema does not exist in the database.
   ```
   so `adapter.get_relation` claims to be able to find the table in the `integrated` database - although it doesn't exist there. And because it doesn't exist - the run fails when it tries to actually read the data from there.

PS that's only a dummy example to visualise the issue, please don't focus on the logic itself

### Environment

```markdown
- OS: macOS Monterey 12.5.1
- Python: 3.9.11
- dbt:
  
  Core:
  - installed: 1.2.1
  - latest:    1.2.1 - Up to date!

  Plugins:
  - redshift: 1.2.1 - Up to date!
  - postgres: 1.2.1 - Up to date!
[...]

Issue #17, opened 2020-04-02
[CT-1616] get_relation method returns None for Redshift external tables


tomekzbrozek:
### Describe the bug
`get_relation` method (https://docs.getdbt.com/docs/writing-code-in-dbt/jinja-context/adapter/#get_relation) returns `None` when provided with database, schema and table name of a Redshift external table (AWS Spectrum). 

### Steps To Reproduce
1. Create an external table with AWS Spectrum: https://docs.aws.amazon.com/redshift/latest/dg/c-spectrum-external-tables.html
2. Attempt to `get_relation` on that external table. 

### Expected behavior
Should return a regular relation as it does for native Redshift tables, e.g. an output like `analytics.spectrum_schema.mytable` instead of `None`

### Screenshots and log output
-

### System information
**Which database are you using dbt with?**
- [ ] postgres
- [x] redshift
- [ ] bigquery
- [ ] snowflake
- [ ] other (specify: - - - )

**The output of `dbt --version`:**
```
0.16.0
```

**The operating system you're using:**
MacOS Mojave 

**The output of `python --version`:**
3.6.2

### Additional context
-

jtcohen6:
Still a bug, still a good first issue for an interested community contributors!

dbeatty10:
Resolved by https://github.com/dbt-labs/dbt-redshift/pull/428

dbeatty10:
This [might not](https://github.com/dbt-labs/dbt-redshift/pull/428#issuecomment-1634232202) actually be resolved by #428 😅

Will either re-open this issue or keep closed depending on the outcome of that conversation.

If it is re-opened, the solution may look similar to https://github.com/dbt-labs/dbt-core/issues/7663, and it may make sense to solve them both at the same time.

trymzet:
FWIW I've adapted [@dbeatty10's solution](https://github.com/dbt-labs/dbt-core/issues/7663#issuecomment-1584854886) for Spectrum and it worked for me (using `dbt-redshift==1.3.0` and `dbt-core==1.3.4`):

```diff
{% macro redshift__list_relations_without_caching(schema_relation) %}

  {% call statement('list_relations_without_caching', fetch_result=True) -%}
    select
        table_catalog as database,
        table_name as name,
        table_schema as schema,
        'table' as type
    from information_schema.tables
    where table_schema ilike '{{ schema_relation.schema }}'
    and table_type = 'BASE TABLE'
    union all
    select
      table_catalog as database,
[...]

PR #428, opened 2023-05-03
Override list_relations_without_caching

jiezhen-chen:
resolves https://github.com/dbt-labs/dbt-redshift/issues/179   resolves https://github.com/dbt-labs/dbt-redshift/issues/217

### Description

Currently, [redshift__list_relations_without_caching](https://github.com/dbt-labs/dbt-redshift/blob/b644963a6bf7d3e64bd71e6ee97a205900089db1/dbt/include/redshift/macros/adapters.sql#L227) leverages [postgres__list_relation_without_caching](https://github.com/dbt-labs/dbt-core/blob/fd7306643fd3a98e3232cbc15bf50eb2f3733603/plugins/postgres/dbt/include/postgres/macros/adapters.sql#L81). Since redshift_connector has its own metadata call([get_tables](https://github.com/aws/amazon-redshift-python-driver/blob/910be71314229e5b2febc60ddf8b8bcc992ea5f0/redshift_connector/cursor.py#L813)) that does the same thing as [postgres__list_relation_without_caching](https://github.com/dbt-labs/dbt-core/blob/fd7306643fd3a98e3232cbc15bf50eb2f3733603/plugins/postgres/dbt/include/postgres/macros/adapters.sql#L81), we are migrating to use the native redshift driver. 

Changes were made to override the method list_relations_without_caching directly, instead of changing the underlying macro sql query. Unit tests were also created under test_redshift_adapter, functional tests created in test_adapter_methods.

### Checklist

    - https://github.com/dbt-labs/docs.getdbt.com/issues/3601


mikealfare:
@jiezhen-chen I like this approach but I have a few questions. They may amount to nothing, but I'd like to walk through it given the shift in method. I'll put them in a review.

dbeatty10:
@jiezhen-chen After a quick peek from a developer experience (DX) perspective, here are two things to address:
1. Issues to close
2. Docs issue for `current_db_only`

### 1. Issues to close
If the merged PR is going to close both #179 and #217, could you update the PR description to say this?
```
resolves #179 resolves #217
```

That way, it will [auto-close both issues](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#:~:text=Use%20full%20syntax%20for%20each%20issue). Otherwise, it will leave #217 open until someone realizes to close it manually.

### 2. Docs issue for `current_db_only`

If the `current_db_only` config item is added, then it will need associated documentation as well.

Could you or the PR review [open an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose)?

tests/functional/adapter/test_adapter_methods.py
```diff
@@ -0,0 +1,125 @@
...
+            "expected.sql": models__expected_sql,
+            "model.sql": models__model_sql,
+            "call_get_relation.sql": models__call_get_relation,
+            "base_view.sql": "{{ config(bind=True) }} select * from {{ ref('model') }}",
+            "get_relation_type.sql": models__get_relation_type,
+            "expected_type.sql": "select 1 as valid_type",
+        }
+
+    @pytest.fixture(scope="class")
+    def project_files(
```
mikealfare, line 90: Does this do anything differently than the `tests` and `models` fixture? I was under the impression that that's what `project` does as long as you call it in a fixture.

tests/functional/adapter/test_adapter_methods.py
```diff
@@ -0,0 +1,125 @@
...
+            "name": "adapter_methods",
+        }
+
+    # snowflake need all tables in CAP name
+    @pytest.fixture(scope="class")
+    def equal_tables(self):
+        return ["call_get_relation", "expected"]
+
+    @pytest.fixture(scope="class")
+    def equal_types(self):
```
mikealfare, line 111: I see `equal_tables` used below, but I don't see where this is used.

tests/functional/adapter/test_adapter_methods.py
```diff
@@ -0,0 +1,125 @@
...
+        write_project_files(project_root, "tests", tests)
+        write_project_files(project_root, "models", models)
+
+    @pytest.fixture(scope="class")
+    def project_config_update(self):
+        return {
+            "name": "adapter_methods",
+        }
+
+    # snowflake need all tables in CAP name
```
dbeatty10, line 105: Is it necessary or desireable to include this comment related to Snowflake?

I'm guessing it is just a copy-paste artifact from [here](https://github.com/dbt-labs/dbt-core/pull/4939/files#diff-a98f96e4919c93a28aee46c131ac9f52bad92403794d7ed03a947c73cb24a96fR93) that can be removed, but wanted to double-check.

```suggestion
```

dbt/adapters/redshift/impl.py
```diff
@@ -114,6 +115,44 @@ def valid_incremental_strategies(self):
     def timestamp_add_sql(self, add_to: str, number: int = 1, interval: str = "hour") -> str:
         return f"{add_to} + interval '{number} {interval}'"
 
+    def _get_cursor(self):
```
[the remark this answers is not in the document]
jiezhen-chen: Yes, you are correct. The _get_cursor is used to call _get_tables. I've instantiated the cursor in  _get_cursor() instead of instantiating a new cursor every time we call the redshift_connector metadata methods. I'm working on overriding other macros with these metadata calls, which will also be using the same  _get_cursor(). That way I don't have to create a new connector.cursor in every one of these new methods.

dbt/adapters/redshift/impl.py
```diff
@@ -114,6 +115,44 @@ def valid_incremental_strategies(self):
...
+    def _get_tables(self, database: Optional[str], schema: Optional[str]):
+        cursor = self._get_cursor()
+        results = []
+        for table in cursor.get_tables(
+            catalog=database,
+            schema_pattern=schema,
+            table_name_pattern=None,
+            types=["VIEW", "TABLE"],
+        ):
+            results.append([table[0], table[1], table[2], table[3]])
```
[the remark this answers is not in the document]
jiezhen-chen: That's a good idea! The return value of the metadata call is just a tuple and indexing is the only method I am aware of. I can create some sort of dictionary as the interface on dbt-redshift though

dbt/include/redshift/macros/adapters.sql
```diff
@@ -224,11 +224,6 @@
 {% endmacro %}
 
 
-{% macro redshift__list_relations_without_caching(schema_relation) %}
```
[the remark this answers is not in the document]
jiezhen-chen: Hmm, good question. I should double check on what it does in the scenario you described. As part of manual testing, I've created a macro in my dbt project that called list_relations_without_caching(), and it did run the method I overrode. But it is a valid point that we should be more careful as I am not completely aware of all the mechanism behind how other pieces in dbt work with macro

dbt/adapters/redshift/connections.py
```diff
@@ -127,6 +131,7 @@ def get_connect_method(self):
             "db_groups": self.credentials.db_groups,
             "region": self.credentials.region,
             "timeout": self.credentials.connect_timeout,
+            "database_metadata_current_db_only": self.credentials.current_db_only,
```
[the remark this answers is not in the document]
jiezhen-chen: Yes that is a redshift_connector parameter and will not be configurable to users.

tests/functional/adapter/test_adapter_methods.py
jiezhen-chen: Yep, good catch!!

dbt/adapters/redshift/impl.py
```diff
@@ -114,6 +115,44 @@ def valid_incremental_strategies(self):
...
+            # kinda feel like it might not be a good thing to do this by index
+        return results
+
+    def list_relations_without_caching(  # type: ignore
+        self, schema_relation: BaseRelation
+    ) -> List[RedshiftRelation]:
+        results = self._get_tables(schema_relation.database, schema_relation.schema)
+        relations = []
+        quote_policy = {"database": True, "schema": True, "identifier": True}
```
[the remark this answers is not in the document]
jiezhen-chen: @mikealfare Not super familiar with QuotePolicy - this part of the code is from the sqladapter [here](https://github.com/dbt-labs/dbt-core/blob/f9abeca231ad20a8d0c4d8154b60bfda48d7fef5/core/dbt/adapters/sql/impl.py#L175)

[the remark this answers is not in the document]
colin-rogers-dbt: I'm not super familiar with this either but my understanding is that `QuotePolicy` should be based on the adapter (i.e. snowflake/spark set their own because they differ from the postgres based default).

dbt/adapters/redshift/impl.py
```diff
@@ -114,6 +115,53 @@ def valid_incremental_strategies(self):
     def timestamp_add_sql(self, add_to: str, number: int = 1, interval: str = "hour") -> str:
         return f"{add_to} + interval '{number} {interval}'"
 
+    def _parse_relation_results(
+        self, database: str, schema: str, name: str, quote_policy: dict[str, bool], type: str
```
colin-rogers-dbt, line 119:
```suggestion
        self, database: str, schema: str, name: str, quote_policy: Dict[str, bool], type: str
```
replace dict (the class) with Dict (the type)

Otherwise you get:
`TypeError: 'type' object is not subscriptable`

dbt/adapters/redshift/impl.py
```diff
@@ -114,6 +115,53 @@ def valid_incremental_strategies(self):
...
+            table_name_pattern=None,
+            types=["VIEW", "TABLE"],
+        ):
+            results.append(
+                {"database": table[0], "schema": table[1], "name": table[2], "type": table[3]}
+            )
+            # kinda feel like it might not be a good thing to do this by index
+        return results
+
+    def list_relations_without_caching(  # type: ignore
```
colin-rogers-dbt, line 152: I think we can just add `@available` to make this available as a macro

jiezhen-chen: Do we want to make this available as a user-callable macro though? Its currently only being used as a macro that supports a user-callable macro such as `get_relation`. I don't believe its available to be called directly by the end user. We could certainly make this change but just double checking that is something we want to do

dataders:
flagging that a related effort to have foreign data wrappers included in the results of `get_relations_without_caching()` surfaced for Postgres itself in https://github.com/dbt-labs/dbt-core/issues/7663

dbt/adapters/redshift/impl.py
colin-rogers-dbt: I believe these are already available as a macro, after some offline conversations I don't think it being a callable macro should break anything.

dbeatty10:
@jiezhen-chen could you share more about the `database_metadata_current_db_only` configuration?

e.g., what it does and why you're including it?

I'm curious if it is it absolutely necessary to add it within this PR or if it feels more like a nice-to-have?

jiezhen-chen:
@dbeatty10 when `database_metadata_current_db_only` is set to false, it allows for cross-database querying, as per this [doc](https://docs.aws.amazon.com/redshift/latest/mgmt/python-configuration-options.html#:~:text=parameter%20is%20optional.-,database_metadata_current_db_only,-Default%20value%20%E2%80%93%20True). This is probably not absolutely necessary for the macro migration, as cross-database querying wasn't supported when we used psycopg2. But would be a nice-to-have and solves https://github.com/dbt-labs/dbt-redshift/issues/179

dbeatty10:
> @dbeatty10 when `database_metadata_current_db_only` is set to false, it allows for cross-database querying, as per this [doc](https://docs.aws.amazon.com/redshift/latest/mgmt/python-configuration-options.html#:~:text=parameter%20is%20optional.-,database_metadata_current_db_only,-Default%20value%20%E2%80%93%20True). This is probably not absolutely necessary for the macro migration, as cross-database querying wasn't supported when we used psycopg2. But would be a nice-to-have and solves #179

Thank you for that info @jiezhen-chen! 🧠

Could we just set `database_metadata_current_db_only` to _always_ be `False` for dbt-redshift?

e.g., What would be the advantages of making it configurable to the end user? Is there any case where they would want it to be `True`?

Unless it is valuable to the end user to be configurable, it would make the documentation and usage more simple to just set it to a constant value that _always_ allows cross-database querying.

jiezhen-chen:
@dbeatty10 There could be performance impact to always setting it to False. We want to provide the option for users to query data from another database, but we don't want to blindly force all users to query from all unconnected databases, since there are potential costs associated with cross-db querying.

dbeatty10:
> What are your thoughts on the default value for `database_metadata_current_db_only`?

Is there a way to dynamically infer whether the user needs multi-db querying and/or datashare support? It would be awesome if we could have _both_ of these without the user needing to add extra configuration themselves:
- support for multi-db querying and datashares
- best performance and lowest cost possible

What would be the performance and/or cost implications to always allowing cross-db querying? e.g., How much extra latency would it add? And how much monetary cost?

colin-rogers-dbt:
> Is there a way to dynamically infer whether the user needs multi-db querying and/or datashare support? 

Since dbt is only creating tables in the target database the only time we would need this set would be if the user wanted to refer to a source that was outside the database. We could have this set based on the source config

mikealfare (approved):

rarup1:
Great that this has been merged into main! @mikealfare @jiezhen-chen which dbt-redshift release will this be available in? We are operating on dbt 1.3. Do you know if I will need to upgrade that to 1.5 before taking advantage of this upgrade?

dbeatty10:
@rarup1 currently, this is slated to require dbt 1.6, which will have its final release in about two weeks.

dbeatty10:
@jiezhen-chen the changelog entry indicated that this issue resolves https://github.com/dbt-labs/dbt-redshift/issues/17, so I added `resolves #17` to formally link this PR as closing that issue. And also added this comment: https://github.com/dbt-labs/dbt-redshift/issues/17#issuecomment-1634194352.

But maybe I shouldn't have done either!

Now that I'm looking at [this implementation](https://github.com/dbt-labs/dbt-redshift/pull/428/files#diff-1cad5abbeecb951c391d29b551b740ff27968260c97f7118bfb865b8abbc9cd5) more closely, I'm not seeing where external tables are covered. I wonder if support for external tables _did_ exist in this PR when [get_tables](https://github.com/aws/amazon-redshift-python-driver/blob/910be71314229e5b2febc60ddf8b8bcc992ea5f0/redshift_connector/cursor.py#L813) was included, but that support was dropped when switching over to `information_schema` views in https://github.com/dbt-labs/dbt-redshift/pull/428/commits/ced9cd100d0721b70a79d3ec80a018a3348011d4?

If this PR does not resolve #17, then we'll want to do a couple follow-up items:
- Remove `resolves #17` from this PR description
- Re-open [#17](https://github.com/dbt-labs/dbt-redshift/issues/17)
- Within a new PR, update the original changelog entry for this PR so it doesn't include [#17](https://github.com/dbt-labs/dbt-redshift/issues/17)

Bonus follow-up items:
- add a note to [#17](https://github.com/dbt-labs/dbt-redshift/pull/17) (and/or [#273](https://github.com/dbt-labs/dbt-redshift/pull/273)) indicating the recommended approach to add support for external tables
- sketch out ideas how to add test cases to cover external tables

jiezhen-chen:
@dbeatty10 You're correct. Because we've pivoted our solution to NOT use [get_tables](https://github.com/aws/amazon-redshift-python-driver/blob/910be71314229e5b2febc60ddf8b8bcc992ea5f0/redshift_connector/cursor.py#L813), this change won't solve issue [17](https://github.com/dbt-labs/dbt-redshift/issues/17) and [217](https://github.com/dbt-labs/dbt-redshift/issues/217).  I'm sorry I didn't update the PR with our latest decision prior to the merge. I'll open a new PR to update the changelog to reflect this. Thanks!

dbeatty10:
No prob @jiezhen-chen. Thank you for updating that changelog entry to reflect the relevant issue in #535, and thanks to @dataders for re-opening #17 and #217 👍

PR #458, opened 2023-05-20
Add autocommit feature to let dbt users run certain db commands in macros

### LINKED ISSUE #425: [ADAP-504] [Regression] cannot run inside a transaction block

### Current Behavior

We have created a basic macro to create database. 

This does not work anymore using the version 1.5

We have the following error message 

```
Encountered an error while running operation: Database Error
  CREATE DATABASE cannot run inside a transaction block
```

### Expected Behavior

The macro should work as worked in the version 1.4

### Steps To Reproduce

Execute the following macro with dbt redshift v1.5
```
{% macro create_ci_database(git_version) %}

{% set create_command %}
    CREATE DATABASE z_ci_{{ git_version }} ;
{% endset %}

{% do run_query(create_command) %}

{{ log("Created database redshift z_ci_" ~ git_version, info=True) }}

{% endmacro %}
```

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt-core: 1.5.0
- dbt-redshift: 1.5.0
```

### Additional Context

_No response_

### LINKED ISSUE #463: [ADAP-573] [Feature] create a target-centric workflow to support autocommit-dependent workflows

### Describe the feature

The redshift connector that underlies this adapter does not have the same transaction semantics as the Postgres connector. Users who want to run certain macros inside of their DBT project are blocked by this changed semantics. 

We want to unblock those users (#425 ). One way that we can do so is letting them to specify autocommit for all connections in their profile config.

For example,

```
profile-to-my-RS-target:
  target: dev
  outputs:
    dev:
      type: redshift
      ...
      [autocommit: False]

profile-to-my-RS-target-with-autocommit-enabled:
  target: dev
  outputs:
    dev:
      type: redshift
      ...
      autocommit: True
```
Users condense specify the autocommit profile with the `--profile` flag on the CLI and in Cloud.

Because autocommit is not something that preserves ACID properties, etc., we need to ensure that the we educate users proper use of this mechanism. 

A docs issue PR exists at https://github.com/dbt-labs/docs.getdbt.com/issues/3401

_No response_

### Who will this benefit?

Redshift users who want to have macros with CREATE DATABASE, VACUUM, and similar.

Yep!

### Anything else?

_No response_

VersusFacit:
resolves #463 

An imperfect but what I think is a necessary solution to unblock users from #425 

## Description of solution

Root cause: Connection objects used in Python must have [`autocommit = True`](https://docs.aws.amazon.com/redshift/latest/mgmt/python-connect-examples.html) to run `CREATE DATABASE`, `VACUUM`, etc..

But, this is a Python-level thing we need to alter. The docs I linked above have this snippet.

image

Yep, this is official guidance. So, we have to alter the connection in the Python post hoc instead of altering the connection instantiation function.

All together -- if we make it an opt-in config for users, we could direct them to define a separate target with `autocommit=True`. This isn't great but it was straightforward enough for me to get 1.5 users unblocked within the sprint!

### Pros of this solution

Let users configure this behavior on their connection profiles. 
* no need for a massive threading rewrite which would be way out of scope of the issue (unblocking users faster)
* finessed opt-in despite the code being kind of blunt ("do this on all connections")
* let users configure on the profile (no env var memorization required for CLI)

------

## The Alternatives Turned Down
1. **Forbid users entirely from using these commands**. Disservice to the user as they worked fine prior to 1.5
2. **Stop using the Redshift connector**. Goes way against our current partner/design philosophy
3. **Enable autocommit for all connections**. Not a great idea: Doing so invites performance hits, data inconsistency faults, no ACID across multiple statements, etc. 
4. **Have dbt silently flip this option on or off for these commands**. My original hope unrealized; requires rewrite of connection contexts.
5. **Okay, then fine. Let's rewrite our connection context code to alter connections based on whatever SQL they're running.** There is no easy way to alter a thread based on what SQL is handed it (thanks Gerda for corroborating that 🙏 ). Neither is this the case for core nor Redshift.

----

## Fix/feature add screenshots to illustrate impact

With this option `autocommit: False` in profile:

image

With this option `autocommit: True` in profile:

image

### 1
Define all macros in a folder with the "autocommit enabled" target
image

### 2
Swap the target on the command line: `--profile TEXT # Which profile to load`. This method is useful for adhoc/scheduled Cloud jobs 💃 

### Followup

1.5 Backport

### Checklist

- [X] I have read [the contributing guide](https://github.com/dbt-labs/dbt-redshift/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [X] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [X] I have run this code in development and it appears to resolve the stated issue
- [X] This PR includes tests, or tests are not required/relevant for this PR
- [X] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR
- [X] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-redshift/blob/main/CONTRIBUTING.md#Adding-CHANGELOG-Entry)

### DIFF
--- a/.changes/unreleased/Fixes-20230520-043039.yaml
+++ b/.changes/unreleased/Fixes-20230520-043039.yaml
@@ -3,4 +3,4 @@ body: Add a new connection param to reenable certain Redshift commands in macros
 time: 2023-05-20T04:30:39.358755-07:00
 custom:
   Author: versusfacit
-  Issue: "425"
+  Issue: "463"
--- a/.changes/unreleased/Fixes-20230520-043039.yaml
+++ b/.changes/unreleased/Fixes-20230520-043039.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Add a new connection param to reenable certain Redshift commands in macros.
+time: 2023-05-20T04:30:39.358755-07:00
+custom:
+  Author: versusfacit
+  Issue: "425"
--- a/tests/functional/adapter/test_autocommit.py
+++ b/tests/functional/adapter/test_autocommit.py
@@ -31,13 +31,14 @@ class TestAutocommitWorksWithTransactionBlocks:
     @pytest.fixture(scope="class")
     def dbt_profile_target(self):
         return {
-            "type": "postgres",
-            "threads": 4,
-            "host": "localhost",
-            "port": int(os.getenv("POSTGRES_TEST_PORT", 5432)),
-            "user": os.getenv("POSTGRES_TEST_USER", "root"),
-            "pass": os.getenv("POSTGRES_TEST_PASS", "password"),
-            "dbname": os.getenv("POSTGRES_TEST_DATABASE", "dbt"),
+            "type": "redshift",
+            "threads": 1,
+            "retries": 6,
+            "host": os.getenv("REDSHIFT_TEST_HOST"),
+            "port": int(os.getenv("REDSHIFT_TEST_PORT")),
+            "user": os.getenv("REDSHIFT_TEST_USER"),
+            "pass": os.getenv("REDSHIFT_TEST_PASS"),
+            "dbname": os.getenv("REDSHIFT_TEST_DBNAME"),
             "autocommit": True,
         }
 
--- a/tests/functional/adapter/test_autocommit.py
+++ b/tests/functional/adapter/test_autocommit.py
@@ -55,6 +55,6 @@ class TestTransactionBlocksPreventCertainCommands:
         return {"macro.sql": _MACROS__CREATE_DB}
 
     def test_normally_create_db_disallowed(self, project):
-        """Monitor if status quo in Cedshift connector changes"""
+        """Monitor if status quo in Redshift connector changes"""
         result, out = run_dbt_and_capture(["run-operation", "create_db_fake"], expect_pass=False)
         assert "CREATE DATABASE cannot run inside a transaction block" in out
--- a/tests/functional/adapter/test_autocommit.py
+++ b/tests/functional/adapter/test_autocommit.py
@@ -1,0 +1,59 @@
+import os
+import pytest
+
+from dbt.tests.util import run_dbt_and_capture
+
+_MACROS__CREATE_DB = """
+{% macro create_db_fake() %}
+
+{% set database = "db_for_test__do_delete_if_you_see_this" %}
+
+{# IF NOT EXISTS not avaiable but Redshift merely returns an error for trying to overwrite #}
+{% set create_command %}
+    CREATE DATABASE {{ database }}
+{% endset %}
+
+{{ log(create_command, info=True) }}
+
+{% do run_query(create_command) %}
+
+{{ log("Created redshift database " ~ database, info=True) }}
+
+{% endmacro %}
+"""
+
+
+class TestAutocommitWorksWithTransactionBlocks:
+    @pytest.fixture(scope="class")
+    def macros(self):
+        return {"macro.sql": _MACROS__CREATE_DB}
+
+    @pytest.fixture(scope="class")
+    def dbt_profile_target(self):
+        return {
+            "type": "postgres",
+            "threads": 4,
+            "host": "localhost",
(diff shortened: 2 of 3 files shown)


mikealfare (approved):
This strikes the right balance and feels like the right place to put the autocommit option. Looks good.

tests/functional/adapter/test_autocommit.py
```diff
@@ -0,0 +1,60 @@
...
+        assert "CREATE DATABASE cannot run inside a transaction block" not in out
+
+
+class TestTransactionBlocksPreventCertainCommands:
+    @pytest.fixture(scope="class")
+    def macros(self):
+        return {"macro.sql": _MACROS__CREATE_DB}
+
+    def test_normally_create_db_disallowed(self, project):
+        """Monitor if status quo in Cedshift connector changes"""
```
dbeatty10, line 58:
```suggestion
        """Monitor if status quo in Redshift connector changes"""
```

dbeatty10 (changes requested):
Would the default behavior in dbt-redshift 1.5 be the same as dbt-redshift 1.4 or would it be different?

From a developer experience perspective, I'd greatly prefer for it to be the same, so I'm going to mark this review as "request changes" until we can agree on the net effect on the developer experience.

Two more questions:
1. Do we know why things worked with `psycopg2` but not `redshift_connector`?
2. What would end users need to do to enable this? Something like this?
    ```yaml
    profile-name:
      target: dev
      outputs:
        dev:
          type: redshift
          ...
          autocommit: True
    ```

VersusFacit:
Hey @dbeatty10 !
- **Would the default behavior in dbt-redshift 1.5 be the same as dbt-redshift 1.4 or would it be different?** Different but forced by the "flags should default to false" constraint. In 1.4, you didn't need to specify this anywhere, but now we need to set it.
- **Do we know why things worked with psychopg2 but not redshift_connector?** No 😕 The RS connector isn't exhaustively documented. Moreover
- **What would end users need to do to enable this? Something like this?** You got it!
profile-to-my-RS-target-with-autocommit-enabled:
  target: dev
  outputs:
    dev:
      type: redshift
      ...
      autocommit: True

McKnight-42 (approved):

tests/functional/adapter/test_autocommit.py
VersusFacit: But I really wanted us to use the Cedshift connector 😭

.changes/unreleased/Fixes-20230520-043039.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Fixes
+body: Add a new connection param to reenable certain Redshift commands in macros.
+time: 2023-05-20T04:30:39.358755-07:00
+custom:
+  Author: versusfacit
+  Issue: "425"
```
dbeatty10, line 6: @VersusFacit just remember to update this to the feature request issue you create:
```suggestion
  Issue: "feature_number_here"
```

dbeatty10 (commented):
Thanks for talking through this in a Slack Huddle with me Mila ❤️

Will approve this to unblock a workaround where users can set `autocommit` within a bespoke profile. This new feature will need to be documented in docs.getdbt.com, so make sure to open a relevant issue.

### Future-facing
I am hoping and wishing we'll discover a way in the future that can allow these commands to work without any intervention from the end user.

The plan we chatted about is that you'll leave https://github.com/dbt-labs/dbt-redshift/issues/425 open and you'll create a separate enhancement request that this PR will resolve instead.

### Checklist of things to do before merging
- [x] create the relevant enhancement request
- [x] update the PR description to change the issue that it resolves
- [x] update the changelog entry to reflect the enhancement request
- [x] [open an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose)

Let me know when these are all ready, and I can give a ✅

dbeatty10 (approved):

PR #466, opened 2023-05-24
Escape `%` symbols in table/view/column comments

### LINKED ISSUE #441: [ADAP-549] [CT-2577] [Regression] Error Only %s and %% are supported in the query
### Is this a regression in a recent version of dbt-core?

- [X] I believe this is a regression in dbt-core functionality
- [X] I have searched the existing issues, and I could not find an existing issue for this regression

### Current Behavior

URLs in markdown files that have a percent sign are causing the error
> Only `%s` and `%%` are supported in the query.

For example this markdown will trigger the error message:
```md
{% docs some_doc %}
Please see this URL:  [Reporting and Analytics Projects](https://somewebsite.com/Reporting%20and%20Analytics%20Projects/AllItems.aspx)
{% enddocs %}
```
Also a single percent sign in the `.md` file will trigger the error.  For example:
```
{% docs some_doc %}
80% of statistics are made up on the spot
{% enddocs %}
```

Also a single percent sign in the sql file will trigger the error.  Such as a percent sign in a comment in the SQL like this:
```sql
select *
from a_table --this is a table where 50% of the records have a null start_date
```

### Expected/Previous Behavior

These errors were not present in version dbt 1.2.

### Steps To Reproduce

Build the dbt project

### Relevant log output

```shell
2023-05-11T22:27:30.313-06:00	[0m04:27:30 [33mRuntime Error in model reporting_and_analytics_projects_init (models/work/initial/sharepoint/reporting_and_analytics_projects_init/reporting_and_analytics_projects_init.sql)[0m

2023-05-11T22:27:30.313-06:00	[0m04:27:30 Only %s and %% are supported in the query.
```

### Environment

```markdown
- OS: debian linux
- Python:Python 3.8.16
- dbt (working version): 1.2.1
- dbt (regression version): 1.5.0
```

redshift

### Additional Context

_No response_

dbeatty10:
resolves #441

### Description

I suspect there's a bug that affects `redshift_connector` (and possibly `pg8000` too) where a [comment](https://docs.aws.amazon.com/redshift/latest/dg/r_COMMENT.html) like the following raises an error regardless if using `paramstyle = 'pyformat'` or `paramstyle = 'format'`:
```sql
  comment on table customers is '95% of customer records';
```

To solve, I took the implementation of `postgres_escape_comment` from https://github.com/dbt-labs/dbt-core/pull/2378 and tweaked it slightly in order to escape all `%` symbols to be `%%` instead.

Added text containing `%` to the functional tests within dbt-core so that we can cover this test case in dbt-redshift and other adapters.

#### Upon opening

#### Before merging
- [x] This PR includes tests
- [x] Docs changes are not required/relevant for this PR
- [x] Revert changes to `dev-requirements.txt` in https://github.com/dbt-labs/dbt-redshift/pull/466/commits/1b41d8c479004194da7eb29ced98577c80f24d0a

### DIFF
--- a/.changes/unreleased/Fixes-20230524-151825.yaml
+++ b/.changes/unreleased/Fixes-20230524-151825.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Escape `%` symbols in table/view/column comments
+time: 2023-05-24T15:18:25.834088-06:00
+custom:
+  Author: dbeatty10
+  Issue: "441"
--- a/dbt/include/redshift/macros/adapters.sql
+++ b/dbt/include/redshift/macros/adapters.sql
@@ -292,3 +292,21 @@
   {% endif %}
 
 {% endmacro %}
+
+{#
+  By using dollar-quoting like this, users can embed anything they want into their comments
+  (including nested dollar-quoting), as long as they do not use this exact dollar-quoting
+  label. It would be nice to just pick a new one but eventually you do have to give up.
+#}
+{% macro postgres_escape_comment(comment) -%}
+  {% if comment is not string %}
+    {% do exceptions.raise_compiler_error('cannot escape a non-string: ' ~ comment) %}
+  {% endif %}
+  {%- set magic = '$dbt_comment_literal_block$' -%}
+  {%- if magic in comment -%}
+    {%- do exceptions.raise_compiler_error('The string ' ~ magic ~ ' is not allowed in comments.') -%}
+  {%- endif -%}
+  {#- -- escape % until the underlying issue is fixed in redshift_connector -#}
+  {%- set comment = comment|replace("%", "%%") -%}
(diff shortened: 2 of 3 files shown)


dbeatty10:
After adding some new text containing `%` to dbt-core, this failed with the same error message as we've seen in bug reports:

image

Flipping the CI from failing to passing gives us confidence that this PR solves _something_ (even if it doesn't solve everything).

image

McKnight-42 (approved):
I like the idea of being able to use historical work to solve similar issue great thinking @dbeatty10 , if we do feel this doesn't solve everything though let's setup some time soon to talk out concerns and options.

McKnight-42 (approved):
I like the idea of being able to use historical work to solve similar issue great thinking @dbeatty10 , if we do feel this doesn't solve everything though let's setup some time soon to talk out concerns and options.

Issue #347, opened 2023-03-01
[CT-2243] [Regression] `dbt seed` doesn't work on certain cases (large row counts)


elongl:
### Is this a regression in a recent version of dbt-redshift?

- [X] I believe this is a regression in dbt-redshift functionality
- [X] I have searched the existing issues, and I could not find an existing issue for this regression

### Current Behavior

I'm not 100% sure when does it happen, but I'm unable to seed CSVs that I previously was able to seed.
I'm getting the following error:

```bash
17:00:52  Finished running 14 seeds, 2 hooks in 0 hours 1 minutes and 22.77 seconds (82.77s).
17:00:52
17:00:52  Completed with 3 errors and 0 warnings:
17:00:52
17:00:52  Runtime Error in seed numeric_column_anomalies_training (data/training/numeric_column_anomalies_training.csv)
17:00:52    'h' format requires -32768 <= number <= 32767
17:00:52
17:00:52  Runtime Error in seed dimension_anomalies_training (data/training/dimension_anomalies_training.csv)
17:00:52    'h' format requires -32768 <= number <= 32767
17:00:52
17:00:52  Runtime Error in seed any_type_column_anomalies_training (data/training/any_type_column_anomalies_training.csv)
17:00:52    'h' format requires -32768 <= number <= 32767
17:00:52
17:00:52  Done. PASS=11 WARN=0 ERROR=3 SKIP=0 TOTAL=14
```

### Expected/Previous Behavior

I should be able to seed them.

### Steps To Reproduce

1. Load [this file](https://github.com/dbt-labs/dbt-redshift/files/10863640/dimension_anomalies_training.csv) to your `seeds` directory.
2. Run `dbt seed`.

Here's [another file](https://github.com/dbt-labs/dbt-redshift/files/10863648/numeric_column_anomalies_training.csv) if needed.

### Environment

```markdown
- OS: macOS
- Python: 3.9.6
- dbt-core (working version): 1.4.0
- dbt-redshift (working version): 1.4.0
- dbt-core (regression version): 1.5.0b2
- dbt-redshift (regression version): 1.5.0b1
```

dbeatty10:
Thanks for discovering and reporting this @elongl 🏆

I was able to confirm this as a regression introduced in dbt-redshift 1.5.0b1. My guess is that it has something to do with us using a new database driver. We migrated from using `psycopg2` to `redshift-connector`in https://github.com/dbt-labs/dbt-redshift/pull/251.

[...]

PR #468, opened 2023-05-24
Use smaller batch size for seeds

dbeatty10:
mitigates (but does not solve root cause of) #347

### Description

For seed CSV files above a certain size, the `redshift_connector` raises the following error:
```
'h' format requires -32768 <= number <= 32767`
```

To mitigate the effect on downstream users, we can lower the default batch size from `10000` to `500`.

In local performance testing using 1.4 (and `psychopg2`), a CSV with 6,000 rows took ~10 seconds to load batch sizes of `10000` to `500` (no difference).

Using 1.5 and (and `redshift_connector`), it took ~13 seconds using a batch size of `500` and ~23 seconds using a batch size of `50`.

### Proposal
My suggestion would be to use a batch size of `500` to mitigate the effect on downstream users while still pursuing a more satisfactory solution (which will probably require a change in `redshift_connector`).

In the meantime, if any users are still running into issues, they can lower their project-specific default even lower with the following file in their dbt project (adjusting the number lower as-needed):
`macros/get_batch_size.sql`
```
{% macro redshift__get_batch_size() %}
  {{ return(500) }}
{% endmacro %}
```

### Testing
I did not include any new CI tests since I was assuming we don't want to add large seed files to CI. This seems okay to me since this PR is not introducing any new behavior -- it is merely override the default batch size for seeds [here](https://github.com/dbt-labs/dbt-core/blob/7c1bd91d0ae3ecea12c7f755cd0b8bb44037f319/core/dbt/include/global_project/macros/materializations/seeds/helpers.sql#L71-L73).

However, I did verify the effect of this change locally.

#### Upon opening

#### Before merging
- [x] I did not include any new CI tests since I was assuming we don't want to add large seed files to CI
- [x] Docs changes are not required/relevant for this PR

### DIFF
--- a/.changes/unreleased/Fixes-20230524-165236.yaml
+++ b/.changes/unreleased/Fixes-20230524-165236.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Use smaller default batch size for seeds
+time: 2023-05-24T16:52:36.915348-06:00
+custom:
+  Author: dbeatty10
+  Issue: "347"
--- a/dbt/include/redshift/macros/materializations/seeds/helpers.sql
+++ b/dbt/include/redshift/macros/materializations/seeds/helpers.sql
@@ -1,0 +1,3 @@
+{% macro redshift__get_batch_size() %}
+  {{ return(500) }}
+{% endmacro %}


McKnight-42 (approved):
this seems like a very fair workaround with minimal increase in time. fantastic writeup seriously helped gaining context so much 👍

dbeatty10:
Thank you @McKnight-42! ❤️ 

This should be backported as well. Can you spearhead the backport process for this one too?

Issue #419, opened 2023-04-27
[ADAP-500] [Bug] dbt-redshift 1.5 does not work with non-standard redshift hostnames


kochalex:
### Is this a new bug in dbt-redshift?

- [X] I believe this is a new bug in dbt-redshift
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

When upgrading to dbt 1.5, I get an 'index out of range error' when attempting to use things that connect to the database.

This is related to using a non-standard hostname for redshift - in my case I am opening a SSH tunnel to Redshift and using localhost as the host.

### Expected Behavior

It works with the same configuration as prior versions of DBT or allows a region configuration to side-step this issue.

### Steps To Reproduce

1. Start a SSH tunnel to redshift (the bug would also manifest if you have a cname to a redshift cluster) and localhost or the cname for the 'host' in your `profiles.yml`.
2. Attempt to run a model or do something that connects to redshift

### Relevant log output

```shell
❯ dbt compile
19:37:05  Running with dbt=1.5.0
19:37:05  Found 111 models, 356 tests, 6 snapshots, 0 analyses, 746 macros, 0 operations, 1 seed file, 135 sources, 0 exposures, 0 metrics, 0 groups
19:37:05
19:37:05  Encountered an error:
list index out of range
19:37:05  Traceback (most recent call last):
  File "/Users/alexkoch/Library/Caches/pypoetry/virtualenvs/espark-dbt-7mBGFPKz-py3.11/lib/python3.11/site-packages/dbt/adapters/redshift/connections.py", line 190, in exception_handler
    yield
  File "/Users/alexkoch/Library/Caches/pypoetry/virtualenvs/espark-dbt-7mBGFPKz-py3.11/lib/python3.11/site-packages/dbt/adapters/sql/connections.py", line 80, in add_query
    cursor = connection.handle.cursor()
             ^^^^^^^^^^^^^^^^^
  File "/Users/alexkoch/Library/Caches/pypoetry/virtualenvs/espark-dbt-7mBGFPKz-py3.11/lib/python3.11/site-packages/dbt/contracts/connection.py", line 94, in handle
    self._handle.resolve(self)
  File "/Users/alexkoch/Library/Caches/pypoetry/virtualenvs/espark-dbt-7mBGFPKz-py3.11/lib/python3.11/site-packages/dbt/contracts/connection.py", line 118, in resolve
    return self.opener(connection)
[...]

PR #485, opened 2023-06-07
Pass `region` param on to redshift conn as is

### LINKED ISSUE #483: [ADAP-609] [CT-2617] [Bug] Database Error   Invalid region provided: net

- [ ] I believe this is a new bug in dbt-core

### Current Behavior

Command: 

    #dbt --no-use-colors  test --project-dir /home/me/projects/pycharm/dbt/dbt/integration_middleware/ --profiles-dir /home/me/projects/pycharm/dbt/dbt/integration_middleware/
    06:29:22  Running with dbt=1.5.1
    06:29:23  Found 31 models, 60 tests, 0 snapshots, 0 analyses, 348 macros, 0 operations, 0 seed files, 0 sources, 0 exposures, 0 metrics, 0 groups
    06:29:23  
    06:29:23  
    06:29:23  Finished running  in 0 hours 0 minutes and 0.03 seconds (0.03s).
    06:29:23  Encountered an error:
    Database Error
      Invalid region provided: net

    
    dbt --version
    Core:
      - installed: 1.5.1
      - latest:    1.5.1 - Up to date!
    
    Plugins:
      - postgres: 1.5.1 - Up to date!
      - redshift: 1.5.2 - Up to date!

This works fine after downgrading the version to:

    dbt --version
    Core:
      - installed: 1.4.6
      - latest:    1.5.1 - Update available!
    
      Your version of dbt-core is out of date!
      You can find instructions for upgrading here:
      https://docs.getdbt.com/docs/installation
    
    Plugins:
      - postgres: 1.4.6 - Update available!
      - redshift: 1.4.0 - Update available!

### Expected Behavior

The dbt tests are run.

### Steps To Reproduce

Upgrade dbt to said version and execute the tests on a Redshift project with the command I stated.
The profiles.yml (originally generated by airbyte) looks like this:

    config:
      partial_parse: true
      printer_width: 120
      send_anonymous_usage_stats: false
      use_colors: true
    normalize:
      outputs:
        prod:
          dbname: MYDB
          host: MYHOST
          pass: MYPASS
          port: MYPORT
          schema: MYSCHEMA
          threads: 1
          type: redshift
          user: MYUSER
      target: prod

### Relevant log output

_No response_

### Environment

```markdown
- OS:Ubuntu 20.04
- Python:3.10.6
- dbt:1.5.1
```

postgres, redshift

### Additional Context

_No response_

### LINKED ISSUE #484: [ADAP-610] [Spike] Investigate if `region` is absolutely required
### Question to answer
- Is passing a `region` to redshift_connector absolutely necessary? i.e., _must_ it be a breaking change in dbt-redshift 1.5?

### Context

1. The `region` was hard-coded to be parsed from the hostname when switching from `psycopg2` to `redshift_connector` in https://github.com/dbt-labs/dbt-redshift/pull/251.
2. An optional `region` config for dbt-redshift profiles was added in https://github.com/dbt-labs/dbt-redshift/pull/420, but it still tries to infer the `region` from the `host` name when it is not provided.

See https://github.com/dbt-labs/dbt-redshift/issues/419 for situations where the `region` can't be inferred from the `host`.

The `redshift_connector` docs have `region` listed as [optional](https://docs.aws.amazon.com/redshift/latest/mgmt/python-configuration-options.html#python-region-option), so maybe there are scenarios in which we can omit supplying it.

### Feedback
https://github.com/dbt-labs/dbt-redshift/issues/419#issuecomment-1578942166

> Looking into this, it seems like region isn't a required parameter at all at least if you are connecting with a username and password. Can we investigate if we can either not require region at all or only require it in situations where it is actually required?

See https://github.com/dbt-labs/dbt-redshift/issues/483 for an additional request to make this non-breaking in 1.5.

colin-rogers-dbt:
resolves #484 resolves #483
### Description
Region is a non-obvious semi-requirement post 1.5 (you don't _need_  to specify it if it's in your hostname), this has been a headache for various users and it's not clear that it's strictly speaking necessary. 

This change makes it a truly optional requirement for dbt-redshift, this removes duplication with the underlying redshift-connector to properly manage and validate this parameter. 

### Checklist

- [x] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose)
    - https://github.com/dbt-labs/docs.getdbt.com/pull/3519


dbt/adapters/redshift/connections.py
```diff
@@ -212,20 +212,23 @@ def get_connect_method(self):
...
 
-            kwargs["region"] = region_value
-
-        # Validate the set region
-        if not _is_valid_region(kwargs["region"]):
+        # Validate the set region if it's available
+        if not kwargs["region"] and not _is_valid_region(kwargs["region"]):
             raise dbt.exceptions.FailedToConnectError(
                 "Invalid region provided: {}".format(kwargs["region"])
             )
```
dbeatty10, line 234: @colin-rogers-dbt I might be reading the logic incorrectly here, but is it supposed to be this?
```python
# Validate the set region if it's available
if kwargs["region"] and not _is_valid_region(kwargs["region"]):
```

dbeatty10:
Loving how this is looking @colin-rogers-dbt 🤩

Since we're removing the region logic entirely in 2dfa187460524f3cc938d7914e14a26881b78684, we'll need to [open an issue to update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose) for this page:
https://docs.getdbt.com/docs/core/connect-data-platform/redshift-setup

Specifically, to make this change in a couple different places:
```diff
-      region: # optional, if not provided, will be determined from host (e.g. host.123.us-east-1.redshift-serverless.amazonaws.com)
+      region: # optional
```

dbeatty10:
As a follow-up to https://github.com/dbt-labs/dbt-redshift/pull/485#issuecomment-1589745306, opened up a PR to update the docs here:
https://github.com/dbt-labs/docs.getdbt.com/pull/3519

alison985:
👋 Hi. A few questions:
1. Will this work for Cloud IDE customers? Both #419 and #483 mention `profiles.yml`, but I don't have one since they don't work with dbt Cloud. I am facing the `Invalid region provided: 0` error with Redshift when moving from dbt v1.2 to dbt v1.5 (latest).
2. Where would I provide region within the dbt Cloud interface?
3. My read of the code in this PR says it's removing inferring the region from the host. If region is a required parameter for the new redshift connection library, `redshift-connector`, then why remove a fallback safety? 
4. If region _is not_ provided to the connection library then what are the risks?

colin-rogers-dbt:
> 1. Will this work for Cloud IDE customers?

This change is specifically for customers (like those using the Cloud IDE) who cannot pass the region parameter at this time. 
> 2. Where would I provide region within the dbt Cloud interface?

For now you will not provide this parameter.

> 3. My read of the code in this PR says it's removing inferring the region from the host. If region is a required parameter for the new redshift connection library, redshift-connector, then why remove a fallback safety?
4. If region is not provided to the connection library then what are the risks?

The region parameter is only required if using IAM auth so for Cloud IDE users this poses no risk. This change has been made after discussion with the maintainers of redshift-connector (where they are also doing this fallback inference in some places) with the decision that we don't want to maintain logic that should be handled by that package.

McKnight-42 (approved):

VersusFacit (commented):
Late approval :p

PR #490, opened 2023-06-13
add merge as a valid incremental strategy

### LINKED ISSUE #402: [ADAP-437] [Feature] Enable "merge" incremental strategy for Redshift

### Describe the feature

Per @kbrock91, we would like to enable Redshift users to use the `merge` strategy for incremental materializations. This is the same thing as https://github.com/dbt-labs/dbt-core/issues/6696 but for Redshift also.

Implementation may be as simple as adding the following to `dbt/adapters/redshift/impl.py`:

```python
    def valid_incremental_strategies(self):
        """The set of standard builtin strategies which this adapter supports out-of-the-box.
        Not used to validate custom strategies defined by end users.
        """
        return ["append", "delete+insert", "merge"]
```

https://github.com/dbt-labs/dbt-redshift/pull/318 is an open PR that is related.

N/A

### Who will this benefit?

Redshift users seeking to use the `merge` strategy for incremental materializations.

N/A

### Anything else?

https://docs.aws.amazon.com/redshift/latest/dg/r_MERGE.html

> This is prerelease documentation for the MERGE command, which is in preview release. The documentation and the feature are both subject to change. We recommend that you use this feature only with test clusters, and not in production environments. Public preview will end on April 5, 2023. Preview clusters and preview serverless workgroups and namespaces will be removed automatically two weeks after the end of the preview. For preview terms and conditions, see Betas and Previews in [AWS Service Terms](https://aws.amazon.com/service-terms/).

jiezhen-chen:
resolves #402

### Description

Add merge as a new incremental strategy for dbt-redshift

### Checklist

- [x] This PR includes tests
- [x] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose)
    - [x] https://github.com/dbt-labs/docs.getdbt.com/issues/3720


dataders:
hi @jiezhen-chen I just opened https://github.com/dbt-labs/docs.getdbt.com/issues/3545 to have this feature documented. We have a bad habit of merging adapter PRs without ensuring the docs are updated. Do you mind drafting a PR to address that issue? Once that PR is done it will be easy to get this merged so that this change can be released next month in dbt-redshift 1.6!

mikealfare:
@jiezhen-chen It looks like unit tests are failing. This looks related to a change that went into `dbt-core` a few weeks ago, where `PartialProject` was updated to require a new argument `dependent_projects_dict`, which can be set equal to `{}`. But these unit tests are complaining about the opposite, that it shouldn't be there. Could you update the `PartialProject` call to omit this argument so we can see if that fixes the issue? We can look into why it changed back later.

mikealfare:
@jiezhen-chen I just heard that we backed out the change that I referenced. I'll update `main` and then refresh your branch, which should resolve the issue.

mikealfare:
> After that, the he main thing to handle would be whether to:
> 1. do some kind of search for DBT_INTERNAL_DEST within [incremental_predicates](https://docs.getdbt.com/docs/build/incremental-models#about-incremental_predicates) and replace with {{ target }}, or
> 2. Give specific guidance to dbt-redshift users to use a different alias than DBT_INTERNAL_DEST
>
> ...
> But all-in-all, Option 1 seems preferable over Option 2.

I think I agree that option 1 is the way to go, without thinking a ton about it. The thing that makes me lean towards it is that I just had to override the table/view materializations for the MV work. Without going into too much detail, the issue was that MVs can't be renamed in Redshift, which means they can't be backed up when replaced by a view/table. This feels like a very similar issue. At some point in the future we'll come back around and break up these materializations a bit, but for now I think we just chalk it up to "Redshift is not Postgres".

jiezhen-chen:
@mikealfare @dbeatty10 Thank you both for all your work on this. I agree that option 1(replace DBT_INTERNAL_DEST with {{ target }}) would be a preferable approach without digging deep into it. I can take up this task to make the changes needed in [merge macro](https://github.com/dbt-labs/dbt-core/blob/5310d3715cdaa109eea03995731425a85cd6cd33/core/dbt/include/global_project/macros/materializations/models/incremental/merge.sql#L4) and hopefully will extend merge support to dbt-redshift

dbeatty10:
@dataders and @jiezhen-chen thanks for pushing up 6e6a116 and 3909713. I am going to test them out locally like this:
```shell
python -m pytest tests/functional/adapter/incremental/test_incremental_predicates.py::TestIncrementalPredicatesMergeRedshift -s
```

When a test fails, `-s` will output the location of a temp directory with `target` output that can be used to debug a little easier locally:
```
tests/functional/adapter/incremental/test_incremental_predicates.py 
=== Test project_root: /private/var/folders/dz/_ffl85j129j7t_djdnkg618c0000gp/T/pytest-of-dbeatty/pytest-437/project0
```

McKnight-42 (approved):

mikealfare (approved):

PR #530, opened 2023-07-13
hard pin to latest compatible version

### LINKED ISSUE #531: [ADAP-699] [Regression] dbt-redshift is suceptible to breaking changes via patch releases to `redshift_connector`
### Is this a regression in a recent version of dbt-redshift?

- [X] I believe this is a regression in dbt-redshift functionality
- [X] I have searched the existing issues, and I could not find an existing issue for this regression

### Current Behavior

Follow up to https://github.com/dbt-labs/dbt-redshift/issues/518

https://github.com/aws/amazon-redshift-python-driver, a package upon which this repo heavily relies, does not follow [Semantic Versioning](https://semver.org/), despite using three-integer release version names. Instead there's weekly release where the patch number is incremented. 

The conventional, Pythonic approach is to compatibility-pin package dependencies to the minor version (`~=x.y.z`), which allows our users to take advantage of any new, non-breaking, bug fixes shipped in patch releases after we release. This is not possible with `redshift_connector`.

Instead, we hard-pin to a specific patch, and only consider incrementing the patch version:
- if there's a user-impacting change in `redshift_connector`, or
- whenever we are ourselves releasing a new major or minor version

### Expected/Previous Behavior

xxx

### Steps To Reproduce

xxx

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt-core (working version):
- dbt-redshift (working version):
- dbt-core (regression version):
- dbt-redshift (regression version):
```

### Additional Context

_No response_

dataders:
resolves #531, #532

### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -85,7 +85,7 @@ setup(
         f"dbt-core~={_core_version()}",
         f"dbt-postgres~={_core_version()}",
         "boto3~=1.26.157",
-        # dbt-core uses these packages deeply, throughout the codebase, and there have been breaking changes in past patch releases (even though these are major-version-one).
+        # dbt-redshift depends deeply on this package. it does not follow SemVer, therefore there have been breaking changes in previous patch releases
         # Pin to the patch or minor version, and bump in each new minor version of dbt-redshift.
         "redshift-connector==2.0.913",
         # installed via dbt-core but referenced directly; don't pin to avoid version conflicts with dbt-core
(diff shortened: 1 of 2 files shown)


mikealfare (approved):

setup.py
```diff
@@ -85,7 +85,9 @@ def _core_version(plugin_version: str = _plugin_version()) -> str:
         f"dbt-core~={_core_version()}",
         f"dbt-postgres~={_core_version()}",
         "boto3~=1.26.26",
-        "redshift-connector~=2.0.911",
+        # dbt-core uses these packages deeply, throughout the codebase, and there have been breaking changes in past patch releases (even though these are major-version-one).
```
mikealfare, line 88: > dbt-core
Technically this package is used by `dbt-redshift`, not `dbt-core`. It's semantics, but should we consider updating this here?

PR #548, opened 2023-07-18
Point `setup.py` to `dbt-core/main` to pick up the alpha

mikealfare:
### Problem

This repo currently has `dbt-core~=1.7.0a1` as a dependency; however, this version of `dbt-core` has not, and will not, be published to PyPI. Hence the dependency is not available when limited to PyPI.

### Solution

Point `setup.py` directly to `dbt-core@main` in the same way `dev-requirements.txt` points to `main`. Remove the now unnecessary entries in `dev-requirements.txt` as well as the version related functions in `setup.py`.

### Checklist


mikealfare:
Not needed

PR #568, opened 2023-08-07
Use the PID to terminate the session

### LINKED ISSUE #553: [ADAP-721] [Bug] [Regression] Canceled or failed run creates a Database Error

### Current Behavior

When dbt tries to cancel a query it raises a Database Error with the pid as output:
```
Encountered an error:
Database Error
  syntax error at or near "123456789"
```
The cases I noticed that are triggering this bug but they may not be exhaustive:
* Cancel a dbt execution while models or tests are running
* A model is failing because of an error while running `--fail-fast` and other running models are cancelled

This is happening only since version 1.5, I don't see the issue on previous versions.

Debugging guess, not tested:
* As per [redshift-connector documentation](https://github.com/aws/amazon-redshift-python-driver#query-using-functions) `fetchone()` seems to return an array instead of a tuple like in psycopg. 
* When a query is cancelled the array is provided instead of the pid, which results in the error. Example: `select pg_terminate_backend([123456789]);`.
https://github.com/dbt-labs/dbt-redshift/blob/cbb45ca585a61b1bb250b48aac69c410f8ef4efb/dbt/adapters/redshift/connections.py#L253

Thanks a lot for your help! I am available if you need more information :)

### Expected Behavior

The query should be cancelled without any error and print usual logs.

### Steps To Reproduce

**Easy steps to reproduce**
1. Use Jaffle Shop project
2. Run the DAG
3. Cancel it while queries are still in `RUN` status (with ctrl+C)
4. Look at the error log

**Harder steps to reproduce**
1. Use a big project which have big models running in parallel (Jaffle Shop project is too simple)
2. Update a model usually running in parallel to make it fail
3. Run the DAG with `--fail-fast`
6. Look at the error log

### Relevant log output

```shell
❯ dbt run 

08:31:16  Running with dbt=1.5.3
08:31:16  Registered adapter: redshift=1.5.8
08:31:17  Found 5 models, 20 tests, 0 snapshots, 0 analyses, 346 macros, 0 operations, 3 seed files, 0 sources, 0 exposures, 0 metrics, 0 groups
08:31:17  
08:31:23  Concurrency: 4 threads (target='dev')
08:31:23  
08:31:23  1 of 5 START sql view model dev.stg_customers .......................... [RUN]
08:31:23  2 of 5 START sql view model dev.stg_orders ............................. [RUN]
08:31:23  3 of 5 START sql view model dev.stg_payments ........................... [RUN]
^C08:31:28  
08:31:28  Finished running  in 0 hours 0 minutes and 10.50 seconds (10.50s).
08:31:28  Encountered an error:
Database Error
  syntax error at or near "1073857832"
08:31:28  3 of 5 ERROR creating sql view model dev.stg_payments .................. [ERROR in 4.89s]
08:31:28  2 of 5 ERROR creating sql view model dev.stg_orders .................... [ERROR in 4.90s]
08:31:28  1 of 5 ERROR creating sql view model dev.stg_customers ................. [ERROR in 4.90s]
```

### Environment

```markdown
- OS: MacOS 13.4.1 Intel
- Python: 3.9.16
- dbt-core: 1.5.3
- dbt-redshift: 1.5.8
```

### Additional Context

_No response_

dbeatty10:
resolves #553
[docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose) N/A

### Problem

When we switched from `psycopg2` to `redshift_connector` in dbt-redshift 1.5.0, the return type of `fetchone()` changed from a `tuple` to an `array`. This caused the query to terminate the session (based on the PID) to not be executed correctly.

### Solution

This solution does two things:
1. Optimistically assumes there is at least one item returned from the PID query and returns that value
2. Directly executes a query to terminate the session (rather than using `self.add_query(sql)` which raises an error for some reason)

One or both can be modified depending on feedback from the code reviewer.

## Also

While troubleshooting the solution noticed that there's a stack trace that didn't happen in dbt-core 1.4.0: https://github.com/dbt-labs/dbt-core/issues/8338

### Checklist

- [ ] 👈 This PR does not include any tests, and obviously none existed previously! Maybe that's okay 🤷
- [x] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc)


jmleon1219:
Can we get this merged?

dbt/adapters/redshift/connections.py
```diff
@@ -240,8 +240,9 @@ class RedshiftConnectionManager(SQLConnectionManager):
     def _get_backend_pid(self):
         sql = "select pg_backend_pid()"
         _, cursor = self.add_query(sql)
+
         res = cursor.fetchone()
-        return res
```
mikealfare, line 244: I think it's safe to assume that we can always get a PID. If we can't, there are probably worse things going on. And to clarify a comment in the PR description, I think what happened is that moving from `psycopg2` to `redshift-connector` changed the behavior in this way:

```
> cursor.fetchone()
> <my_pid>
```
```
> cursor.fetchone()
> (<my_pid>,)
```

It's worth noting that the latter is [PEP-249](https://peps.python.org/pep-0249/#fetchone) compliant. This would align with other issues we've seen where `psycopg2` is not PEP-249 compliant but `redshift-connector` is.

tl;dr; The way you have it is fine, I'm just providing context.

dbt/adapters/redshift/connections.py
```diff
@@ -253,9 +254,8 @@ def cancel(self, connection: Connection):
             raise
 
         sql = f"select pg_terminate_backend({pid})"
-        _, cursor = self.add_query(sql)
-        res = cursor.fetchone()
-        logger.debug(f"Cancel query '{connection.name}': {res}")
+        cursor = connection.handle.cursor()
+        cursor.execute(sql)
```
mikealfare, line 258: This feels like the right way to handle this anyway; if we're cancelling a query, we're probably already in a failed state and don't necessarily need results from the database. However, can we keep the log event in here? Maybe make it something like:

```python
logger.debug(f"Cancel query on: '{connection.name}' with PID: {pid}")
```

dbt/adapters/redshift/connections.py
```diff
@@ -253,9 +254,8 @@ def cancel(self, connection: Connection):
             raise
 
         sql = f"select pg_terminate_backend({pid})"
-        _, cursor = self.add_query(sql)
```
mikealfare, line 256: This probably started throwing an error because Redshift may not let you execute `pg_terminate_backend` as part of a multi-statement execution. We started seeing issues with how multi-statement executions within transactions were handled in `psycopg2` versus `redshift-connector`. Again it came down to `psycopg2` not being PEP-249 compliant whereas `redshift-connector` is. Unfortunately we built each adapter around `psycopg2`, which leaves some artifacts like this one to work out.

mikealfare (commented):
Most of the review is context. The one change I would recommend is adding the logging statement backend. It doesn't need to be verbatim what I put, but I imagine that could be useful for troubleshooting when things go awry.

dbt/adapters/redshift/connections.py
dbeatty10: 🧠 Thanks for this context

dbt/adapters/redshift/connections.py
dbeatty10, line 258: Debug-level logging for both the high-level message as well as the actual SQL being executed:
```suggestion
        logger.debug(f"Cancel query on: '{connection.name}' with PID: {pid}")
        logger.debug(sql)
        cursor.execute(sql)
```

mikealfare: It looks like going from `add_query` to `execute` causes this to be dropped from the observed call stack in the mock. The unit tests fails because it doesn't see the call from `execute`. From looking at the unit test you updated below, it seems like you might just need to mock `execute` as well (but also in the other tests where it failed).

dbeatty10: Removed it from the expected call stack in https://github.com/dbt-labs/dbt-redshift/pull/568/commits/a36d7f89dfcd12b23d56f3518cb96ca59a0b0de5, so passing the remaining tests.

But that particular call is no longer asserted within the unit tests now. Do you want to leave it as-is, or would you prefer to try to add a mock for `execute` as well?

mikealfare: Oh funny, for some reason I thought it was more than one test. I guess it was just he different os/py versions. Looks good to me then.

mikealfare (approved):

PR #569, opened 2023-08-08
ADAP-761: Add retry logic to flaky MV tests to avoid "cannot open relation with OID" error

mikealfare:
### Problem

Materialized view tests are sporadically failing due to a "cannot open relation with OID" error. This error can occur when there are multiple objects being created and dropped in parallel, as happens with these integration tests. The appears to be caused by the fact that queries on the metadata tables are not locked like other queries, hence a record may exist on those tables that corresponds to an object that is in the process of being dropped.

### Solution

Add retry logic that will retry the `run_dbt_and_capture` command up to ten times, until it returns successfully. If it does not return successfully, then it will return `None`.

This fix will need to be backported to `1.6.latest` as it's affecting CI on both `main` and `1.6.latest`.

### Checklist


tests/functional/adapter/materialized_view_tests/utils.py
```diff
@@ -67,3 +68,20 @@ def query_autorefresh(project, relation: RedshiftRelation) -> bool:
         and trim(mv.db_name) ilike '{ relation.database }'
     """
     return project.run_sql(sql, fetch="one")[0]
+
+
+def run_dbt_and_capture_with_retries(args: List[str], max_retries: int = 10):
```
colin-rogers-dbt, line 73: since we're adding this in the mv scope it'll probably help folks coming along later if the function name indicates it's defined here and not in core

tests/functional/adapter/materialized_view_tests/utils.py
```diff
@@ -67,3 +68,20 @@ def query_autorefresh(project, relation: RedshiftRelation) -> bool:
...
+    """
+    We need to retry `run_dbt` calls on Redshift because we get sporadic failures of the form:
+
+        Database Error in model my_materialized_view (models/my_materialized_view.sql)
+        could not open relation with OID 14957412
+    """
+    retries = 0
+    while retries < max_retries:
+        try:
+            # there's no point to using this with expect_pass=False
```
colin-rogers-dbt, line 83: should we add a `sleep` here? Just thinking if we run these in parallel are we just going to be spamming the DWH

tests/functional/adapter/materialized_view_tests/utils.py
mikealfare: When you say "here", are you concerned more about Redshift, MVs, or both?

tests/functional/adapter/materialized_view_tests/utils.py
mikealfare: The tests run in parallel, but this loop should be in sequence. Would adding sleep do anything to solve for the parallel tests?

tests/functional/adapter/materialized_view_tests/utils.py
colin-rogers-dbt: hmm I guess technically it _should_ be both but that seems kinda unwieldy, how about:
```suggestion
def mv_run_dbt_and_capture_with_retries(args: List[str], max_retries: int = 10):
```

tests/functional/adapter/materialized_view_tests/utils.py
colin-rogers-dbt: ...sigh not really

colin-rogers-dbt (approved):
nit on function name but otherwise LGTM

mikealfare:
@cla-bot check

PR #574, opened 2023-08-10
ADAP-761: [backport] Add retry logic to flaky MV tests to avoid "cannot open relation with OID" error

mikealfare:
See #569


nathaniel-may (approved):

mikealfare:
Re-opening to trigger stalled CI

PR #583, opened 2023-08-15
Auto-refresh on MVs not working as expected

### LINKED ISSUE #560: [ADAP-738] [Bug] Auto-refresh on MVs not working as expected

### Current Behavior

When I set auto_refresh to false in the v1.6, it's still set to true in the compiled and executed code. 

image

and then in subsequent runs, I hit upon a permissions error 

image (1)

### Expected Behavior

When I execute my model as a MV with auto-refresh set to false, auto refresh should equal no in the compiled code. Also in subsequent runs, if nothing has changed (because I used the `apply` on_configuration_change, it should `continue`. 

### Steps To Reproduce

1. Create a dbt Cloud account with a dev env set to 1.6 patch version 
2. Create a dbt model
3. Add the configuration:
```
{{
config(
  materialized='materialized_vew',
  auto_refresh = 'false',
  on_configuration_change = 'apply',
)
}}
```
4. run dbt build --select <model_name>
5. run dbt build --select <model_name> again to see the permission error

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt-core:
- dbt-redshift:
```

### Additional Context

_No response_

McKnight-42:
resolves #560 

### Problem
auto_refresh changes are not reflected properly in mv on rerun if auto_refresh value is a str

### Solution
allow for bool and str values to be accepted and just turn it internally to a corresponding boolean value.

### Checklist


tests/functional/adapter/materialized_view_tests/test_materialized_views.py
```diff
@@ -108,10 +108,26 @@ def change_config_via_alter(project, materialized_view):
         new_model = initial_model.replace("dist='id',", "dist='id', auto_refresh=True")
         set_model_file(project, materialized_view, new_model)
 
+    @staticmethod
+    def change_config_via_alter_str_true(project, materialized_view):
```
McKnight-42, line 112: We might be able to make some slight improvements in this method and the `check_state_alter_is_applied` method. Possibly by just making the original method take in the third keyword arg, we can change in the tests. I tried doing it with f-strings but it didn't seem to work. any thoughts on this @mikealfare?

McKnight-42:
@cla-bot check

dbt/adapters/redshift/relation_configs/materialized_view.py
```diff
@@ -123,9 +123,25 @@ def parse_model_node(cls, model_node: ModelNode) -> dict:
...
-            "autorefresh": model_node.config.extra.get("auto_refresh"),
         }
 
+        autorefresh_value = model_node.config.extra.get("auto_refresh")
+        if autorefresh_value is not None:
+            if isinstance(autorefresh_value, bool):
+                config_dict["autorefresh"] = autorefresh_value
+            elif isinstance(autorefresh_value, str):
+                lower_autorefresh = autorefresh_value.lower()
+                if lower_autorefresh == "true":
```
colin-rogers-dbt, line 134: It _feels_ like this should be a utility function shared across core, especially as it looks like in other places we're relying on the soon to be deprecated `distutils`: https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/task/test.py#L54

mikealfare: That's fair. In fact, this PR is necessary specifically because we made the decision that we should make it easier on the user by always accepting string versions of boolean configs across the board, not just for `auto_refresh`. Since this needs to be backported to 1.6, are you comfortable with an adapter-specific solution here with a general solution for 1.7 and forward?

tests/functional/adapter/materialized_view_tests/test_materialized_views.py
mikealfare: I think what may be complicating this is trying to test the various inputs via the change monitoring tests. Have you considered a more focused test? We effectively want to test different model config inputs; so a class that creates a bunch of new models with different settings for `auto_refresh` would suffice.

dbt/adapters/redshift/relation_configs/materialized_view.py
colin-rogers-dbt: makes sense to me

colin-rogers-dbt (approved):

PR #590, opened 2023-08-29
ADAP-841: Reformat macro file structure to be more modular and organized

mikealfare:
### Problem

Implementing `replace` functionality requires macros that work across relations, but which are implemented at the relation grain. This requires more modularity and organization for our macros.

### Solution

In order to make the updates easier to review, we will break out the pure reformatting from the actual code changes. This PR only moves macros around to fit a more granular file structure. The macros themselves are not altered; code is not altered and signatures are not altered.

### Checklist


dbt/include/redshift/macros/materializations/materialized_view.sql
mikealfare, line 1: Move all of these into their own files in `macros/relations/materialized_view/*.sql`

dbt/include/redshift/macros/adapters.sql
```diff
@@ -312,12 +312,3 @@
   {% endif %}
 
 {% endmacro %}
-
-
-{% macro redshift__get_drop_relation_sql(relation) %}
```
mikealfare, line 317: Move this into it's own file in `macros/relations/drop.sql`.

dbt/include/redshift/macros/relations/materialized_view/_replace.sql
dataders, line 1: why is this filename underscored prefixed?

dbt/include/redshift/macros/materializations/materialized_view.sql
```diff
@@ -1,106 +1,5 @@
...
-    {{ redshift__get_drop_relation_sql(existing_relation) }};
-    {{ get_create_materialized_view_as_sql(relation, sql) }}
-{% endmacro %}
-
-
 {% macro redshift__get_materialized_view_configuration_changes(existing_relation, new_config) %}
     {% set _existing_materialized_view = redshift__describe_materialized_view(existing_relation) %}
     {% set _configuration_changes = existing_relation.materialized_view_config_changeset(_existing_materialized_view, new_config) %}
     {% do return(_configuration_changes) %}
 {% endmacro %}
```
dataders, line 5: why does this macro remain here in `adapters.sql`?

dbt/include/redshift/macros/relations/drop.sql
```diff
@@ -0,0 +1,7 @@
+{% macro redshift__get_drop_relation_sql(relation) %}
+    {%- if relation.is_materialized_view -%}
+        {{ redshift__drop_materialized_view(relation) }}
+    {%- else -%}
+        drop {{ relation.type }} if exists {{ relation }} cascade
+    {%- endif -%}
+{% endmacro %}
```
dataders, line 7: when these changes roll out to other materializations, we'll likely see more macros called within here.. right?

dataders (commented):
I have some small questions, mostly to build and confirm my understanding.

Most importantly, imo, if the purpose of this PR is to make the actual refactor PR's changes easier to see, I'd like to see that follow-on PR, before merging this one. I'm imagining that this could be done by building a new branch off of this PR's `feature/materialized-views/adap-841`.

dbt/include/redshift/macros/relations/materialized_view/drop.sql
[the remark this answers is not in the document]
mikealfare: Probably not, that's kind of by design. The idea is to make `relations/<relation_type>/<operation>.sql` look a bit like the docs page for the database platform. That was a decision that a few of us on the team came to in order to make the jinja easier to navigate. Does this cause you concern?

dbt/include/redshift/macros/relations/materialized_view/_replace.sql
mikealfare: Replace at the relation-specific grain doesn't make sense; that's actually the crux of the change. Most macros have a single relation as an argument. Even with something like `rename`, the second relation is really just a name, not the object. But `replace` inherently has two relations (the one being replaced, and the target that will be there afterwards). And most importantly, these two relations need not be the same type. Currently we support materialized views replacing materialized view. We have raised this to the relation-agnostic layer in `dbt-core` and supplied relation-specific components that go into each of the adapters. `replace` in `dbt-core` could be overridden, but I would recommend against it. As such, this relation-specific, adapter-specific version will be deleted as part of the feature work in the next PR.

tl;dr; This is going away, but moved for the sake of following along.

dbt/include/redshift/macros/relations/drop.sql
mikealfare: Ideally this gets standardized into `dbt-core`. I'll look back at the git blame, but I believe this was a renamed version of drop_relation that I then added materialized view to as an option. But we really should audit where this used and use the version that's in `dbt-core`.

In general, yes, we will see more relations (so table and view) show up in these "switchboard" macros (e.g. `get_drop_sql`) and more calls to them as well, with less calls to relation-specific macros (e.g. `get_drop_table_sql`). For this MV work I am trying to touch as few macros/files as possible as the sheer number of files affected in the last approach caused heartburn. At a bare minimum, I need all operations for MVs and drop/rename for all relation types (so I can replace tables and views with MVs and vice versa).

mikealfare:
> I have some small questions, mostly to build and confirm my understanding.
> 
> Most importantly, imo, if the purpose of this PR is to make the actual refactor PR's changes easier to see, I'd like to see that follow-on PR, before merging this one. I'm imagining that this could be done by building a new branch off of this PR's `feature/materialized-views/adap-841`.

That's what we did for `dbt-snowflake`. You can see the two PRs here:
https://github.com/dbt-labs/dbt-snowflake/pull/751
https://github.com/dbt-labs/dbt-snowflake/pull/753

We determined this was a better approach than putting it all into the same PR. We made that decision by basically putting it all in the same PR initially and then confusing people. Either way, all of these changes are going into `main`; it's just a matter of whether we break it up.

dbt/include/redshift/macros/materializations/materialized_view.sql
dataders: whoops i read the file heading wrong. i see it's actually in `dbt/include/redshift/macros/materializations/materialized_view.sql` not `adapters.sql`

dbt/include/redshift/macros/relations/materialized_view/drop.sql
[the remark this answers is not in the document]
dataders: all good just confirming!

colin-rogers-dbt (approved):

PR #603, opened 2023-09-13
add relation macros and add relation renameable/replaceable config

colin-rogers-dbt:
resolves #

### Checklist


.changes/unreleased/Fixes-20230912-133327.yaml
```diff
@@ -1,5 +1,5 @@
 kind: Fixes
```
mikealfare, line 1: We're not considering this a bug since we're not backporting it. We're thinking of it as a new feature (the ability to swap MVs with table/view and vice versa). The body would also probably change given that context.

dbt/include/redshift/macros/relations/table/replace.sql
```diff
@@ -0,0 +1,47 @@
...
+          validator=validation.any[list, basestring]) -%}
+  {%- set sql_header = config.get('sql_header', none) -%}
+  {%- set backup = config.get('backup') -%}
+
+  {{ sql_header if sql_header is not none }}
+
+  {%- set contract_config = config.get('contract') -%}
+  {%- if contract_config.enforced -%}
+
+  create table
```
mikealfare, line 18: These (this line and line 36) should be `create or replace` instead of just `create`.

dev-requirements.txt
```diff
@@ -1,8 +1,8 @@
 # install latest changes in dbt-core + dbt-postgres
 # TODO: how to switch from HEAD to x.y.latest branches after minor releases?
-git+https://github.com/dbt-labs/dbt-core.git#egg=dbt-core&subdirectory=core
-git+https://github.com/dbt-labs/dbt-core.git#egg=dbt-tests-adapter&subdirectory=tests/adapter
-git+https://github.com/dbt-labs/dbt-core.git#egg=dbt-postgres&subdirectory=plugins/postgres
+git+https://github.com/dbt-labs/dbt-core.git@fixReplaceSQLReplaced#egg=dbt-core&subdirectory=core
```
mikealfare, line 3: This is merged in `dbt-core`, so this can be reverted back to `main` again.

tests/functional/adapter/materialized_view_tests/test_materialized_views.py
```diff
@@ -70,15 +70,9 @@ def test_materialized_view_create_idempotent(self, project, my_materialized_view
         )
         assert self.query_relation_type(project, my_materialized_view) == "materialized_view"
 
-    @pytest.mark.skip(
-        "The current implementation does not support overwriting materialized views with tables."
-    )
     def test_table_replaces_materialized_view(self, project, my_materialized_view):
```
mikealfare, line 73: These tests can be removed entirely since they're inheriting from `dbt-core`. Ultimately if everything is implemented, then these test classes should just have configuration and no test cases.

dbt/include/redshift/macros/relations/table/replace.sql
colin-rogers-dbt: I don't think redshift supports `or replace`

mikealfare: If Redshift doesn't support `create or replace` for tables, then we need to remove `RelationType.Table` from `replaceable_relations` and then delete this file.

colin-rogers-dbt: good point! duh

mikealfare (approved):

PR #630, opened 2023-10-11
adding IAM role authentication support for redshift

### LINKED ISSUE #623: [ADAP-926] [Feature] IAM authentication - make 'user' an optional property

### Describe the feature

Redshift supports IAM authentication without a redshift user name via [GetClusterCredentialsWithIAM](https://docs.aws.amazon.com/redshift/latest/APIReference/API_GetClusterCredentialsWithIAM.html), which

> Returns a database user name and temporary password with temporary authorization to log in to an Amazon Redshift database. The database user is mapped 1:1 to the source AWS Identity and Access Management (IAM) identity.

It would be awesome if dbt-redshift supported this authentication method as well, allowing for `user` to be optional configuration when using `method: iam` for authentication.

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

abbywh:
Committer: Abby Whittier <abbywhittier309@gmail.com>

resolves #623

### Problem

dbt-redshift currently does not natively support the redshift authentication method [get-cluster-credentials-with-iam](https://docs.aws.amazon.com/cli/latest/reference/redshift/get-cluster-credentials-with-iam.html). A dbt customer using IAM roles as db users currently has no way to authenticate as those users. 

### Solution

There is now a new redshift connection method to authenticate as an IAM role, called IAMR. This strongly resembles the classic IAM authentication, except the user field is unnecessary. 

### Checklist


abbywh:
Is this being tested now? Any chance we can reopen for tracking purposes?

dbt/adapters/redshift/connections.py
```diff
@@ -104,9 +105,9 @@ def parse(cls, user_sslmode: UserSSLMode) -> "RedshiftSSLConfig":
 @dataclass
 class RedshiftCredentials(Credentials):
     host: str
-    user: str
     port: Port
     method: str = RedshiftConnectionMethod.DATABASE  # type: ignore
+    user: Optional[str] = None  # type: ignore
```
colin-rogers-dbt, line 110: if we're making this optional we'll need to add a validation that the user supplied this value for connection methods that require it (i.e. `RedshiftConnectionMethod.DATABASE`)

colin-rogers-dbt:
@abbywh looks like there's something wrong with your git setup and is throwing off our CLA check

abbywh:
@colin-rogers-dbt do you have any documentation for your bot? I can explain what happened. I ammended the commit to remove my private email address in favor of my public, and you can see my email in the commit, but it's still blocking. I would happily parse through how to fix it (maybe a git filter repo?) if I can find some documentation around it.

dbt/adapters/redshift/connections.py
abbywh: agreed, will work on this and update the code.

abbywh:
for anyone digging this up, make sure you update the config as given by the bot, then do `git amend --reset-author` as opposed to `git amend --author="Author Name <email@address.com>"`.

dataders:
@jiezhen-chen @soksamnanglim @sathiish-kumar y'all wanna take a look here as well?

jiezhen-chen:
Instead of introducing a new `IAMR` auth method, would yall consider having it under `IAM`? Looks like the behavior of `IAM` and `IAMR` are the same except `user` is now optional, can we optionally pass in `self.credential.user` only when it is provided in the yml file?

abbywh:
@jiezhen-chen i also considered that option. would be happy to change it to that, but i was worried about muddling the error cases for the iam authentication. 

for example, consider i'm a user that wants to use IAM auth and omits the user field. the error i would get would be something like: `Error: IAM role xyz does not have permission GetClusterCredentialWithIam on cluster 123`. This might incorrectly lead a user to add that permission. We also couldn't catch that error and throw a different message, because then we wouldn't cover the error case where a user actually intends to use iamr and had that permission misconfigured. 

I think this is mostly a stylistic choice that should be left to the maintainers. But we should add a note about that error message to the documentation if we choose to combine the routes. Just let me know which you would prefer knowing this and I will implement it 😃

jiezhen-chen:
@abbywh Having a IAMR auth method makes sense. I think that if you want to use GetClusterCredentialsWithIAM v2 where IAM role is used, you'd also need to set `group_federation` to `true`, according to this [issue](https://github.com/aws/amazon-redshift-python-driver/issues/189#issuecomment-1734226917). I don't see this added as a parameter passed into the connect() method. Is the current IAMR connect() method able to connect using GetClusterCredentialsWithIAM v2 without needing to set `group_federation` to `true`?

abbywh:
yes you are correct there was a mistake in my testing setup. I will make the changes to the tests that should catch this + that code change.

misteliy:
Hi all, I would like to understand the progress of this issue since we would like to switch completely to IAM (not specifying user explicitly but rather through IAM) - thus IAMR would be great to have as soon as possible.

abbywh:
+1. This not being deployed is still causing big issues for us.

joaobernardopa:
Hello everyone, this would actually be a game changer for us as in our current setup we still need to create all external schemas...

saratvysyaraju:
Hey All,

Any updates on when this feature may get shipped? Our team is currently waiting for this to begin adopting the IAM Auth

mikealfare:
@abbywh I'll be looking at this PR over the next week or so, with the intent of merging it within the next two weeks. Given the duration for which this has been open, I wanted to check with you to see if you're available to make updates, if needed. I don't mind taking it over, and I would do so in a way that you maintain credit for your contributions either way. But we prefer to offer the contributor the opportunity to complete the PR if so desired.

mikealfare:
I pushed the merge conflict resolution so that the branch can pull updates from `main`. It's a large update because this PR pre-dates the core/adapter de-coupling work; I think my IDE said 94 files or so.

mikealfare (commented):
I left a few comments, mostly around aesthetics and issues arising from the de-coupling work.

The one thing that we'll need to add is an integration test verifying that we can connect with an IAM role. I need to setup some infra for that test to run in CI. In the meantime, we need to do a few things:

- add `REDSHIFT_TEST_IAM_ROLE` and `REDSHIFT_TEST_IAM_PROFILE` to `./test.env.example` [here](https://github.com/abbywh/dbt-redshift/blob/Feature/iamr_auth/test.env.example)
- pull these new values into the test fixtures in `./tests/conftest.py` [here](https://github.com/abbywh/dbt-redshift/blob/d574ef42d19701fc4171c363379db0f6603fbe84/tests/conftest.py#L13-L22)
- create a test module for the new tests: `./tests/functional/adapter/test_auth_methods.py`
- create a test in the new module that connects via IAM role and runs a basic query to demonstrate access

The test should probably remove things like `user` for a positive test. And it's probably worth adding a negative test as well, by removing `role`.

mikealfare:
In the interest of time, I will take over this PR in https://github.com/dbt-labs/dbt-redshift/pull/781 and close this one.

PR #637, opened 2023-10-19
ADAP-835: Optimize manual refresh on auto-refreshed materialized views

mikealfare:
resolves dbt-labs/dbt-core#6911

### Problem

`dbt run` will issue a manual refresh of a materialized view if there are no changes detected during a run. This happens regardless of whether the materialized view is setup to be automatically refreshed. If the materialized view is already scheduled to be refreshed, there is no need to insert an additional refresh. This will add cost and potentially introduce unexpected data as the refresh schedule will not match what's set with the extra refresh.

### Solution

dbt will inspect the materialized view to determine if it is set to automatically refresh. If it is, and there are no changes, then `dbt run` will do nothing. If it is not set to automatically refresh, then `dbt run` will issue a refresh command.

### Checklist


mikealfare:
@cla-bot check

PR #647, opened 2023-10-26
Fixes #641 by using the svv_mv_info view which is queryable

### LINKED ISSUE #641: [ADAP-959] [Bug] macro describe_materialized_view does not work with redshift serverless

### Current Behavior

When doing a `dbt run` on a redshift serverless cluster, permission denied errors will happen when dbt attempts to query the svt_mv_info table. This is due to that table not being queryable by any user in redshift serverless [as described here](https://docs.aws.amazon.com/redshift/latest/mgmt/serverless-monitoring.html#serverless_views-monitoring).

This can be remedied by a custom macro for `redshift__describe_materialized_view`, which references the `svv_mv_info` view instead of the `svt_mv_info` table as shown below

```
{% macro redshift__describe_materialized_view(relation) %}
    {#-
        These need to be separate queries because redshift will not let you run queries
        against svv_table_info and pg_views in the same query. The same is true of svv_redshift_columns.
    -#}

    {%- set _materialized_view_sql -%}
        select
            tb.database,
            tb.schema,
            tb.table,
            tb.diststyle,
            tb.sortkey1,
            mv.autorefresh
        from svv_table_info tb
        left join svv_mv_info mv
            on mv.database_name = tb.database
            and mv.schema_name = tb.schema
            and mv.name = tb.table
        where tb.table ilike '{{ relation.identifier }}'
        and tb.schema ilike '{{ relation.schema }}'
        and tb.database ilike '{{ relation.database }}'
    {%- endset %}
    {% set _materialized_view = run_query(_materialized_view_sql) %}

    {%- set _query_sql -%}
        select
            vw.definition
        from pg_views vw
        where vw.viewname = '{{ relation.identifier }}'
        and vw.schemaname = '{{ relation.schema }}'
        and vw.definition ilike '%create materialized view%'
    {%- endset %}
    {% set _query = run_query(_query_sql) %}

    {% do return({'materialized_view': _materialized_view, 'query': _query}) %}

{% endmacro %}
```

### Expected Behavior

properly introspect the materialized view in redshift serverless

### Steps To Reproduce

`dbt run` with a model with materialization=materialized_view on a redshift serverless cluster

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python: 
- dbt-core: 1.6.2
- dbt-redshift:1.6.2
```

### Additional Context

_No response_

reptillicus:
resolves #641

### Problem
When doing a dbt run on a redshift serverless cluster, permission denied errors will happen when dbt attempts to query the svt_mv_info table. This is due to that table not being queryable by any user in redshift serverless [as described here](https://docs.aws.amazon.com/redshift/latest/mgmt/serverless-monitoring.html#serverless_views-monitoring).

### Solution
This can be remedied by a custom macro for redshift__describe_materialized_view, which references the svv_mv_info view instead of the svt_mv_info table as shown below

### Checklist

### DIFF
--- a/dbt/include/redshift/macros/relations/materialized_view/describe.sql
+++ b/dbt/include/redshift/macros/relations/materialized_view/describe.sql
@@ -13,6 +13,7 @@
             tb.sortkey1,
             mv.autorefresh
         from svv_table_info tb
+        -- svv_mv_info is queryable by Redshift Serverless, but stv_mv_info is not
         left join svv_mv_info mv
             on mv.database_name = tb.database
             and mv.schema_name = tb.schema
--- a/dbt/include/redshift/macros/relations/materialized_view/describe.sql
+++ b/dbt/include/redshift/macros/relations/materialized_view/describe.sql
@@ -13,9 +13,9 @@
             tb.sortkey1,
             mv.autorefresh
         from svv_table_info tb
-        left join stv_mv_info mv
-            on mv.db_name = tb.database
-            and mv.schema = tb.schema
+        left join svv_mv_info mv
+            on mv.database_name = tb.database
+            and mv.schema_name = tb.schema
             and mv.name = tb.table
         where tb.table ilike '{{ relation.identifier }}'
         and tb.schema ilike '{{ relation.schema }}'
--- a/.changes/unreleased/Fixes-20231026-164623.yaml
+++ b/.changes/unreleased/Fixes-20231026-164623.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Fix describe_materialized_view for Redshift Serverless
+time: 2023-10-26T16:46:23.253837-06:00
+custom:
+  Author: reptillicus
+  Issue: "641"


reptillicus:
@dbeatty10 here's a wee little PR

colin-rogers-dbt:
@jiezhen-chen any reason we shouldn't use `svv_mv_info`?

mikealfare:
> @jiezhen-chen any reason we shouldn't use `svv_mv_info`?

In particular, is there a reason to use `svv_mv_info` over `stv_mv_info` or vice versa? We want to make sure we're not trading one problem for another. On the surface these seem like very similar objects. I'd also like to stick a comment somewhere that explains why we're using one over the other so we don't accidentally revert this in the future.

dbt/include/redshift/macros/relations/materialized_view/describe.sql
```diff
@@ -13,9 +13,9 @@
             tb.sortkey1,
             mv.autorefresh
         from svv_table_info tb
-        left join stv_mv_info mv
-            on mv.db_name = tb.database
-            and mv.schema = tb.schema
+        left join svv_mv_info mv
```
dbeatty10, line 16: To address @mikealfare's suggestion [here](https://github.com/dbt-labs/dbt-redshift/pull/647#issuecomment-1782202473).

@mikealfare did you want an explicit link to the GitHub issue as well? If so, you just add a code suggestion of what you are looking for?

```suggestion
        -- svv_mv_info is queryable by Redshift Serverless, but svt_mv_info is not
        left join svv_mv_info mv
```

mikealfare: I'm an idiot, I just didn't read. @jiezhen-chen provided the rationale in the summary above. If I still get to be nit-picky, your suggested change has a typo; 'stv_mv_info` is misspelled. The comment addresses my concern though.

dbeatty10: Good eyes! Nit-picky is appreciated. I fixed the spelling above, and I'm going to commit this suggestion momentarily.

mikealfare (approved):

PR #653, opened 2023-10-31
ADAP-973: Fix issue where materialized views were being mapped as views in catalog queries

### LINKED ISSUE #652: [ADAP-973] [Bug] Materialized views show up as regular views in catalog queries

### Current Behavior

Materialized views show up as regular views in catalog queries.

### Expected Behavior

Materialized views should show up as materialized views in catalog queries.

### Steps To Reproduce

1. Create a materialized view
2. Run `docs generate`
3. Notice the type for the materialized view is "VIEW"

See https://github.com/dbt-labs/dbt-core/issues/8864 for more details.

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt-core:
- dbt-redshift:
```

### Additional Context

_No response_

mikealfare:
resolves #652, dbt-labs/dbt-core#8864

### Problem

Materialized views show up as views in catalog queries.

### Solution

- write a test demonstrating the issue
- update the catalog query to correctly assign type for materialized views

### Checklist

### DIFF
--- a/dbt/include/redshift/macros/catalog.sql
+++ b/dbt/include/redshift/macros/catalog.sql
@@ -43,7 +43,7 @@
             sch.nspname as table_schema,
             tbl.relname as table_name,
             case
-                when materialized_views.relname is not null then 'MATERIALIZED VIEW'
+                when tbl.relkind = 'v' and materialized_views.relname is not null then 'MATERIALIZED VIEW'
                 when tbl.relkind = 'v' then 'VIEW'
                 else 'BASE TABLE'
             end as table_type,
--- a/dbt/include/redshift/macros/catalog.sql
+++ b/dbt/include/redshift/macros/catalog.sql
@@ -23,13 +23,28 @@
         order by "column_index"
     ),
 
+    materialized_views as (
+        select
+            table_schema  as nspname,
+            table_name    as relname
+        from information_schema.views
+        where (
+            {%- for schema in schemas -%}
+                upper(table_schema) = upper('{{ schema }}'){%- if not loop.last %} or {% endif -%}
+            {%- endfor -%}
+        )
+        and table_catalog = '{{ database }}'
+        and view_definition ilike '%create materialized view%'
+    ),
+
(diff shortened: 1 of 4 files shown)


dbt/include/redshift/macros/catalog.sql
```diff
@@ -23,13 +23,28 @@
...
+            table_schema  as nspname,
+            table_name    as relname
+        from information_schema.views
+        where (
+            {%- for schema in schemas -%}
+                upper(table_schema) = upper('{{ schema }}'){%- if not loop.last %} or {% endif -%}
+            {%- endfor -%}
+        )
+        and table_catalog = '{{ database }}'
+        and view_definition ilike '%create materialized view%'
```
VersusFacit, line 37: This will be performance intensive. Is there a faster (regex?) alternative?

VersusFacit (commented):
One nit. Otherwise, straightforward. May want to leave a comment about the `case`-`when` being a short circuit operation though. The important of that precedence of checking the view is materialized first could be lost without that context somewhere.

dbt/include/redshift/macros/catalog.sql
mikealfare: Oh, perhaps. But we're already doing this in `list_relations_without_caching`. To my knowledge, there's no good way to do this check.

mikealfare:
> One nit. Otherwise, straightforward. May want to live a comment about the `case`-`when` being a short circuit operation though. The important of that precedence of checking the view is materialized first could be lost without that context somewhere.

That's a good point. I chose instead to add the condition in the case statement. Let me know what you think.

colin-rogers-dbt (approved):

dbt/include/redshift/macros/catalog.sql
VersusFacit: Hmmm, fair enough. I appreciate the point about maintaining state of the art for now.

VersusFacit (approved):

PR #674, opened 2023-11-29
Fix #624 by removing unnecessary boto3 version peg

### LINKED ISSUE #624: [ADAP-927] [Feature] Update boto3 to 1.29.* version

### Describe the feature

Boto3 is pinned to the 1.26.* series, which is three major versions behind. This prevents use of updated awscli, boto3, botocore in conjunction with dbt-redshift

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

hexDoor:
resolves #624

### Problem
The `boto3` dependency is as of writing, pegged 7 major versions behind the latest `boto3` release.
The only context where `boto3` is referenced within the `dbt-redshift` repository is within a unit test which raises the question on whether the `boto3` version peg is even necessary since it is already a transitive dependency of `redshift-connector`.
At the current time, this version peg frustrates compatibility of `dbt-redshift` with the latest `awscli`, `boto3` and `botocore` versions as well as other unrelated dbt plugins.

### Solution
Remove `boto3` as a direct dependency for `dbt-redshift`

### Checklist

### DIFF
--- a/.changes/unreleased/Dependencies-20231130-044332.yaml
+++ b/.changes/unreleased/Dependencies-20231130-044332.yaml
@@ -1,0 +1,6 @@
+kind: Dependencies
+body: Remove direct boto3 dependency
+time: 2023-11-30T04:43:32.872452+11:00
+custom:
+  Author: hexDoor
+  PR: "674"
--- a/setup.py
+++ b/setup.py
@@ -84,7 +84,6 @@ setup(
     install_requires=[
         f"dbt-core~={_core_version()}",
         f"dbt-postgres~={_core_version()}",
-        "boto3~=1.26.157",
         # dbt-redshift depends deeply on this package. it does not follow SemVer, therefore there have been breaking changes in previous patch releases
         # Pin to the patch or minor version, and bump in each new minor version of dbt-redshift.
         "redshift-connector==2.0.915",
--- a/tests/functional/adapter/test_basic.py
+++ b/tests/functional/adapter/test_basic.py
@@ -92,6 +92,7 @@ class TestBaseAdapterMethod(BaseAdapterMethod):
     pass
 
 
+@pytest.mark.skip(reason="Known flakey test to be reviewed")
 class TestDocsGenerateRedshift(BaseDocsGenerate):
     @pytest.fixture(scope="class")
     def expected_catalog(self, project, profile_user):
--- a/tests/unit/test_redshift_adapter.py
+++ b/tests/unit/test_redshift_adapter.py
@@ -144,7 +144,6 @@ class TestRedshiftAdapter(unittest.TestCase):
         )
 
     @mock.patch("redshift_connector.connect", Mock())
-    @mock.patch("boto3.Session", Mock())
     def test_explicit_iam_conn_with_profile(self):
         self.config.credentials = self.config.credentials.replace(
             method="iam",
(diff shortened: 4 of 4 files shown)


hexDoor:
CLA has now been signed.

mikealfare:
@cla-bot check

mikealfare:
@hexDoor Do you mind adding a changelog entry (`changie new`)?

@dbt-labs/core-adapters How do folks feel about this? I would normally say to at least keep the dependency in the requirements, unpinned, if we use it directly. But given the limited usage and the fact that we're likely going to have `redshift-connector` here for the foreseeable future, I think it's alright to remove it entirely. This was probably a holdover from before we started using `redshift-connector` and should have been removed at that point.

VersusFacit:
I'm pretty anxious about unintended consequences here. Sure, we aren't using this directly but what about this being update by other packages? Indeed, we've got failures here. We need to ensure this change is self-contained before moving forward.

hexDoor:
> Sure, we aren't using this directly but what about this being update by other packages?

@VersusFacit Could I confirm what you mean by `this`?
If we're talking about how other packages are using `dbt-redshift`, I would be very concerned for the dependency management of those packages if they were using `boto3` in their package without declaring a direct dependency.
With that in mind, I would say with reasonable certainty that removing this dependency is not going to affect end users who are following best practices.

As for the test failure (`tests/functional/adapter/test_basic.py::TestDocsGenerateRedshift::test_run_and_generate`), I believe it is unrelated to this change and might be resolved with a rerun.
```
17:50:14  Encountered an error while generating catalog: Database Error
  could not open relation with OID 22039866
17:50:14  dbt encountered 1 failure while writing the catalog
17:50:14  Catalog written to /tmp/pytest-of-runner/pytest-0/popen-gw3/project3/target/catalog.json
```

hexDoor:
> Having worked in this space for a few years now, I take absolutely nothing for granted as to the dependency management of Python packages. We've had a lot of incidents over the years where things move out from underneath us and break unceremoniously. Several version pins are main frankly to protect us at times. Hence, my desire for caution and verification.

I see. I would love to avoid causing an incident myself (especially on the public stage) 😆
However, would love to see some RCA's or discussions on some of these incidents out of interest.

> I'll give things a rerun just in case you're right.

I don't think the Adapter Integration Test workflow got re-run as I can't see a new entry in the list 😞

mikealfare:
> As for the test failure (tests/functional/adapter/test_basic.py::TestDocsGenerateRedshift::test_run_and_generate), I believe it is unrelated to this change and might be resolved with a rerun.

Just to remove this from the discourse, this test is a known flaky test. It fails routinely, necessitating a re-run. It is on our backlog to look into and resolve. It is unrelated to this change.

VersusFacit:
@hexDoor 

We talked about this internally a little bit. We feel comfortable with merging this PR. Before we do so though, can we get you to do one more commit? It'd be great to do the following. Transparently, it's to deal with flaky tests and to illuminate references to this package entirely where they are unneeded.

- [ ] remove each `@mock.patch("boto3.Session", Mock())` found in `test/unit/test_redshift_adapter.py`, that's like 6 lines I think
- [ ] remove `tests/functional/adapter/test_basic.py::TestDocsGenerateRedshift::test_run_and_generate` to get your PR consistently green

That will get us out of stale test hell and resolve this issue at once :D

VersusFacit (approved):
Hooray, we're green. Thanks for doing us a solid @hexDoor and for being so responsive. Happy to see this merged.

PR #689, opened 2023-12-13
Migrate references to base adapter 

VersusFacit:


dbt/adapters/redshift/connections.py
```diff
@@ -33,7 +34,7 @@ def get_message(self) -> str:
 logger = AdapterLogger("Redshift")
 
 
-drop_lock: Lock = dbt.flags.MP_CONTEXT.Lock()  # type: ignore
+drop_lock: Lock = dbt.mp_context._MP_CONTEXT.Lock()  # type: ignore
```
VersusFacit, line 37: remove this

dbt/adapters/redshift/impl.py
```diff
@@ -96,7 +96,7 @@ def verify_database(self, database):
         ra3_node = self.config.credentials.ra3_node
 
         if database.lower() != expected.lower() and not ra3_node:
-            raise dbt.exceptions.NotImplementedError(
+            raise dbt.common.exceptions.NotImplementedError(
```
mikealfare, line 99: Can we use the built-in `NotImplementedError`?

dbt/adapters/redshift/relation.py
```diff
@@ -60,31 +59,28 @@ def relation_max_name_length(self):
         return MAX_CHARACTERS_IN_IDENTIFIER
 
     @classmethod
-    def from_runtime_config(cls, runtime_config: RuntimeConfigObject) -> RelationConfigBase:
-        model_node: ModelNode = runtime_config.model
-        relation_type: str = model_node.config.materialized
+    def from_config(cls, config: RelationConfig) -> RelationConfigBase:
```
mikealfare, line 62: Should this be `from_relation_config`? It looks like we went with that in other places.

mikealfare: We can resolve this in the future if needed. It's a naming convention concern, not a functionality concern.

mikealfare (approved):

PR #694, opened 2023-12-19
Implementation of metadata-based freshness

### LINKED ISSUE #615: [ADAP-911] Support Metadata Freshness
### Describe the feature

Support metadata-based freshness by implementing the new macro and feature flag described in [#7012](https://github.com/dbt-labs/dbt-core/issues/7012).

### Who will this benefit?

Everyone who wants faster freshness results from Redshift!

mikealfare:
resolves #615

### Problem

The current implementation of source freshness requires querying the data and requires the user provide a datetime field. This is slower and more expensive than it should be. It doesn't scale across multiple models. And some models do not have an appropriate datetime field.

### Solution

Use source metadata where available.

### Checklist


dbt/include/redshift/macros/metadata/relation_last_modified.sql
```diff
@@ -0,0 +1,29 @@
+{% macro redshift__get_relation_last_modified(information_schema, relations) -%}
+
+    {%- call statement('last_modified', fetch_result=True) -%}
+        select
+            ns.nspname as "schema",
+            c.relname as identifier,
+            max(qd.start_time) as last_modified,
```
mikealfare, line 7: Is there a reason to use `end_time` instead of `start_time`?

dbt/include/redshift/macros/metadata/relation_last_modified.sql
```diff
@@ -0,0 +1,29 @@
...
+            {{ current_timestamp() }} as snapshotted_at
+        from pg_class c
+        join pg_namespace ns
+            on ns.oid = c.relnamespace
+        join sys_query_detail qd
+            on qd.table_id = c.oid
+        where qd.step_name in ('insert', 'merge', 'mergejoin')
```
mikealfare, line 14: Does this cover all scenarios?

colin-rogers-dbt: hmm I actually wonder if 'merge' and 'mergejoin' should be excluded: https://docs.aws.amazon.com/redshift/latest/dg/r_EXPLAIN.html

Looks like the only "step" that modifies a table is 'insert'

mikealfare: Good catch, that's what I get for assuming. I thought that was referencing a "merge" into an existing table, like with `pandas`.

dbt/include/redshift/macros/metadata/relation_last_modified.sql
colin-rogers-dbt: I suppose because the start_time better reflects when the data being inserted was actually generated? (i.e. if an insert takes 2 hours the end_time would make the data look 2 hours "fresher" than it is)

colin-rogers-dbt: in most cases they probably won't be that different tho

mikealfare: Yeah, this was my rationale as well. I'll keep it as is then.

PR #711, opened 2024-02-06
[Bug] Initialize sqlparse lexer prior to using sqlparse

### LINKED ISSUE #710: [Bug] Runtime Error  'Lexer' object has no attribute '_SQL_REGEX'

### Current Behavior

dbt Cloud users experiencing an intermittent runtime error in random job runs which stems from the `sqlparse` library that dbt uses([link](https://github.com/andialbrecht/sqlparse/blob/f101546dafa921edfea5b3107731504665b758ea/sqlparse/lexer.py#L133)). But this appears to only be affecting Redshift connected projects. Issue was also reported on Discourse [here](https://discourse.getdbt.com/t/runtime-error-lexer-object-has-no-attribute-sql-regex/11557).

Some notes from the internal correspondence with Core team:
- `sqlparse` is a library that we use to combine the SQL from ephemeral models with the non-ephemeral model SQL.  It looks like this error would be an even more edge-casey failure than the one that we addressed in the [referenced ticket](https://github.com/dbt-labs/dbt-core/pull/8215).
- This is likely a variation on the original issue, which is that the sqlparse lexer isn't initialized when we use it to parse the SQL. They set the "default_instance" before it's initialized so another thread checking if the default instance is set in that tiny tiny window will see that it is, but it hasn't been initialized.  In order to fix this we'd probably need to write an internal version of their initialization code and pin the code. Or vendor it.
- Looks like dbt-redshift directly uses sqlparse but just for splitting.  It's the only adapter that does it though and it looks like it went in as of 1.5 so lines up. It also (on main) no longer does probably with dbt-adapter being pulled out ([link](https://github.com/dbt-labs/dbt-redshift/blob/74de320b6aa216b6c57cdb2b1304e4121f642038/dbt/adapters/redshift/connections.py#L353)).

### Expected Behavior

Job runs not failing with the aforementioned error.

### Steps To Reproduce

There's no specific step to reproduce this as it happens on random intervals and so far the only consistent thing found in the behaviour has been the `dbt-redshift` adapter. Error has occurred on dbt version 1.5, 1.6 and 1.7. 

### Relevant log output

```shell
an error:
Runtime Error
  'Lexer' object has no attribute '_SQL_REGEX'
```

### Environment

```markdown
- OS:
- Python:
- core: Running with dbt=1.6.9
-  adapter: redshift=1.6.7
```

### Additional Context

_No response_

mikealfare:
resolves #710

### Problem

We are seeing an error when running `dbt docs generate` on a set of sources that uses external tables. This error is similar to what we saw here: https://github.com/dbt-labs/dbt-core/issues/8217.

### Solution

We need to initialize `Lexer` in `sqlparse` prior to using `sqlparse`, as is done in this PR: https://github.com/dbt-labs/dbt-core/pull/8215

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20240206-132326.yaml
+++ b/.changes/unreleased/Fixes-20240206-132326.yaml
@@ -1,0 +1,7 @@
+kind: Fixes
+body: Initialize sqlparse.Lexer to resolve issue with `dbt docs generate` that includes
+  external tables
+time: 2024-02-06T13:23:26.061133-05:00
+custom:
+  Author: mikealfare
+  Issue: "710"
--- a/dbt/adapters/redshift/connections.py
+++ b/dbt/adapters/redshift/connections.py
@@ -354,6 +354,7 @@ class RedshiftConnectionManager(SQLConnectionManager):
         connection = None
         cursor = None
 
+        self._initialize_sqlparse_lexer()
         queries = sqlparse.split(sql)
 
         for query in queries:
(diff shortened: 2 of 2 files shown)


QMalcolm (approved):
Looks great! Thank you for doing this ❤️

colin-rogers-dbt (approved):

PR #719, opened 2024-02-24
Fix the renamed relations code against 1.8

### LINKED ISSUE #693: [ADAP-1077] [Drops Table In Separate Transaction]  - Tables are dropped then the _tmp table renamed, causing other queries referencing the table to fail

### Current Behavior

Models with table materialization are dropped then the transaction is committed.  This causes queries referencing the refreshing model to fail if they run at the same time since the table doesn't exist.

I've tried setting `autocommit: true ` in profiles.yml to no effect.

### Expected Behavior

The old behavior was to drop/rename in one transaction.

### Steps To Reproduce

Run `dbt run` command in dbt-redshift >=1.7.1 (core >=1.7.3) against a table model.  

dbt-redshift 1.6.5 (core 1.6.9)  runs in a single transaction

### Relevant log output

```shell
[0m08:54:14.597104 [debug] [Thread-2  ]: On model.$$$$: BEGIN
[0m08:54:14.646101 [debug] [Thread-2  ]: SQL status: SUCCESS in 0.0 seconds
[0m08:54:14.669096 [debug] [Thread-2  ]: Applying DROP to: $$$$
[0m08:54:14.671270 [debug] [Thread-2  ]: Using redshift connection "model.$$$$"
[0m08:54:14.672171 [debug] [Thread-2  ]: On model.$$$$: /* {"app": "dbt", "dbt_version": "1.7.3", "profile_name": "redshift", "target_name": "$$$$", "node_id": "model.$$$$"} */
drop table if exists $$$$ cascade
[0m08:54:14.719695 [debug] [Thread-2  ]: SQL status: SUCCESS in 0.0 seconds
[0m08:54:14.723391 [debug] [Thread-2  ]: On model.$$$$: COMMIT
```

### Environment

```markdown
- OS: macOS sonoma
- Python: 3.8.13
- dbt-core: 1.7.4
- dbt-redshift: 1.7.1
```

### Additional Context

_No response_

VersusFacit:
resolves #693 

### Problem

~~There are too many BEGINs and COMMITs in the logs. This is due to how we're nesting macros. This is a problem when drops and renames need to operate together in a single transaction.~~

This is a red herring. The real problem is `renamed_relations` is not evaluating right on instantiation and causing drops where there shouldn't be any.

### Solution
Change instantiation semantics in [core](https://github.com/dbt-labs/dbt-core/pull/9682) and here. Add test.

### Checklist


dbt/adapters/redshift/relation.py
mikealfare, line 1: I don't think we need to make this update for the fix. I think we just need to make the update in `dbt-core`/`dbt-adapters` so that this override takes.

VersusFacit: It fails without it.

dev-requirements.txt
```diff
@@ -2,7 +2,7 @@
 # TODO: how to switch from HEAD to x.y.latest branches after minor releases?
 git+https://github.com/dbt-labs/dbt-postgres.git@main
 git+https://github.com/dbt-labs/dbt-core.git#egg=dbt-core&subdirectory=core
-git+https://github.com/dbt-labs/dbt-adapters.git
+git+https://github.com/dbt-labs/dbt-adapters.git@ADAP-108/fix_renamed_relations_for_1.8
```
VersusFacit, line 5: Need to change this to 1.7

mikealfare:
This looks like it might be a dupe. I'll come back after merging the other PR and clean this up if that's the case.

mikealfare:
After fixing `rename_relations`, this PR has been reduced to the changelog, hence I'm closing it as a dupe. Please refer to https://github.com/dbt-labs/dbt-core/issues/9681 for more information.

PR #728, opened 2024-03-02
Add explicit dependency on sqlparse that used to inherit from core

mikealfare:


.github/workflows/main.yml
```diff
@@ -59,7 +59,6 @@ jobs:
           python -m pip --version
           pre-commit --version
           mypy --version
-          dbt --version
```
mikealfare, line 62: The only intention of this step is to ensure `dbt-core` is installed, which is no longer required, whereas the other instances are meant to verify that `dbt-redshift` is a functioning python package.

PR #731, opened 2024-03-17
[Bug] Python driver 2.0.918 causes SSL error. Need to revert back to 2.0.917

### LINKED ISSUE #730: [Bug] Python driver 2.0.918 causes SSL error. Need to revert back to 2.0.917

### Current Behavior

There is a bug in 2.0.918 which we have updated the dbt-redshift dependencies to (including in dbt Cloud) that causes connecting to RA3 nodes with data sharing to fail with a "Database Error Server refuses SSL" error message.

### Expected Behavior

This should not fail this way. If we revert back to 2.0.917, there should be no impact (per the Redshift team). 

### Steps To Reproduce

1. Pip install redshift driver to point to 2.0.918
2. Connect to a RA3 node with the preview data sharing functionality on
3. run dbt debg

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt-core: 1.7+
- dbt-redshift: 1.7+
```

### Additional Context

_No response_

mikealfare:
resolves #730

### Problem

A bug causing an SSL error was introduced to `redshift-connector` in `2.0.918`.

### Solution

We need to pin `redshift-connector` to `<2.0.917` to resolve the issue for now. We will need to determine a long term solution, which may be as simple as skipping `2.0.918`, as we did with `2.0.914`, after `redshift-connector` resolves the underlying issue.

This will need to be backported to `1.7.latest`, where it was introduced.

### Checklist

### DIFF
--- a/.changes/unreleased/Fixes-20240317-113447.yaml
+++ b/.changes/unreleased/Fixes-20240317-113447.yaml
@@ -1,0 +1,6 @@
+kind: Fixes
+body: Pin `redshift-connector` to <2.0.918 to avoid SSL error introduced in 2.0.918
+time: 2024-03-17T11:34:47.873169-04:00
+custom:
+  Author: mikealfare
+  Issue: "730"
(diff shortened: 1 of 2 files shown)


VersusFacit (approved):
Don't merge till you get it green 😎

amychen1776:
@mikealfare was this released?I don't see it in the changelog.

mikealfare:
> @mikealfare was this released?I don't see it in the changelog.

Nope, nope it wasn't. Not sure how we missed that.

PR #732, opened 2024-03-18
Add Redshift internal build process workflow

VersusFacit:
resolves #38 in epic

### Problem

We need a workflow to release things internally

### Solution

Write the workflow. I'll be opening up a PR for the workflow this one calls.

Followup actions including reconfiguring secrets and doing a full run start to finish!

### Checklist

- [X] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [X] I have run this code in development and it appears to resolve the stated issue
- [X] This PR includes tests, or tests are not required/relevant for this PR
- [X] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc) or this PR has already received feedback and approval from Product or DX

### DIFF
--- a/.changes/unreleased/Features-20240318-030018.yaml
+++ b/.changes/unreleased/Features-20240318-030018.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Add new workflow for internal patch releases
+time: 2024-03-18T03:00:18.845618-07:00
+custom:
+  Author: versusfacit
+  Issue: "38"
--- a/.github/workflows/release-internal.yml
+++ b/.github/workflows/release-internal.yml
@@ -16,49 +16,24 @@ on:
         type: string
         default: "python -c \"import dbt.adapters.redshift\""
         required: true
-      dbms_name:
-        description: "The name of the warehouse the adapter connects to."
-        type: string
-        default: "redshift"
-        required: true
-  workflow_call:
-    inputs:
-      version_number:
-        description: "The release version number (i.e. 1.0.0b1)"
-        type: string
-        required: true
-      sha:
-        description: "The sha to use (leave empty to use latest on main)"
-        type: string
-        required: false
-      package_test_command:
-        description: "Package test command"
-        type: string
-        default: "python -c \"import dbt.adapters.redshift\""
-        required: true
-      dbms_name:
-        description: "The name of the warehouse the adapter connects to."
-        type: string
-        default: "redshift"
-        required: true
 
 defaults:
   run:
-    shell: bash
+    shell: "bash"
 
 env:
   PYTHON_TARGET_VERSION: 3.11
 
 jobs:
   invoke-reusable-workflow:
-    name: Build and Release Internally
+    name: "Build and Release Internally"
 
-    uses: VersusFacit/dbt-release/.github/workflows/internal-archive-release.yml@main
+    uses: "VersusFacit/dbt-release/.github/workflows/internal-archive-release.yml@main"
 
     with:
-      version_number: ${{ inputs.version_number }}
-      package_test_command: ${{ inputs.package_test_command }}
-      dbms_name: ${{ inputs.dbms_name }}
-      sha: ${{ inputs.sha }}
+      version_number: "${{ inputs.version_number }}"
+      package_test_command: "${{ inputs.package_test_command }}"
+      dbms_name: "redshift"
+      sha: "${{ inputs.sha }}"
 
-    secrets: inherit
+    secrets: "inherit"
--- a/.github/workflows/release-internal.yml
+++ b/.github/workflows/release-internal.yml
@@ -1,18 +1,3 @@
-# **what?**
-# Release workflow provides the following steps:
-# - checkout the given commit
-# - validate version in sources and changelog file for given version
-# - bump the version and generate a changelog if needed
-# - merge all changes to the target branch if needed
-# - run unit and integration tests against given commit
-# - build and package that SHA
-# - release it to internal repo
-#
-# **why?**
-# Ensure an automated and tested release process
-#
-# **when?**
-# This workflow can be run manually on demand or can be called by other workflows
 name: Release internal patch
 
 on:
--- a/.github/workflows/release-internal.yml
+++ b/.github/workflows/release-internal.yml
@@ -114,6 +114,12 @@ jobs:
     needs: [create-internal-release]
 
     steps:
+      - name: "Install Python Dependencies"
+        run: |
+          python -m pip install --user --upgrade pip
+          python -m pip install --upgrade setuptools wheel twine check-wheel-contents
+          python -m pip --version
+
       - name: "Configure AWS profile for upload"
         run: |
           aws configure set aws_access_key_id ${{ secrets.AWS_ACCESS_KEY_ID }} --profile ${{ secrets.AWS_PROFILE_NAME }}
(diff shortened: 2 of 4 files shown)


setup.py
```diff
@@ -50,7 +60,7 @@ def _plugin_version() -> str:
     install_requires=[
         "dbt-common>=0.1.0a1,<2.0",
         "dbt-adapters>=0.1.0a1,<2.0",
-        f"dbt-postgres~={_plugin_version()}",
+        f"dbt-postgres~={_plugin_version_trim()}",
         # dbt-redshift depends deeply on this package. it does not follow SemVer, therefore there have been breaking changes in previous patch releases
```
VersusFacit, line 64: This disclaimer holds here

.changes/unreleased/Features-20240318-030018.yaml
emmyoop, line 1: I know core generally does not have changelogs for changes in `/.github` since they do not affect users.

.github/workflows/release-internal.yml
```diff
@@ -0,0 +1,64 @@
+name: Release internal patch
```
emmyoop, line 1: Please add the what/why/when documentation to the top of the workflow.

.github/workflows/release-internal.yml
```diff
@@ -0,0 +1,64 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      version_number:
+        description: "The release version number (i.e. 1.0.0b1)"
+        type: string
+        required: true
+      sha:
+        description: "The sha to use (leave empty to use latest on main)"
```
emmyoop, line 11: I would make this a required field and default it to `main` if the expectation is that we will mostly run this off `main`.  It makes manually submissions more explicit.  Maybe rename this input to "sha_or_main" to make it really clear in the dropdown.

.github/workflows/release-internal.yml
```diff
@@ -0,0 +1,64 @@
...
+      package_test_command:
+        description: "Package test command"
+        type: string
+        default: "python -c \"import dbt.adapters.redshift\""
+        required: true
+      dbms_name:
+        description: "The name of the warehouse the adapter connects to."
+        type: string
+        default: "redshift"
+        required: true
```
emmyoop, line 23: Since this will never _not_ be redshift, drop this as an input and hardcode it into the workflow as an env var.

.github/workflows/release-internal.yml
```diff
@@ -0,0 +1,64 @@
...
+      sha:
+        description: "The sha to use (leave empty to use latest on main)"
+        type: string
+        required: false
+      package_test_command:
+        description: "Package test command"
+        type: string
+        default: "python -c \"import dbt.adapters.redshift\""
+        required: true
```
emmyoop, line 18: Will we realistically ever want to use any other test command?  If not, drop this as an input and make it an env var.  The less inputs there can be, the easier this is to submit.

.github/workflows/release-internal.yml
```diff
@@ -0,0 +1,64 @@
...
+      dbms_name:
+        description: "The name of the warehouse the adapter connects to."
+        type: string
+        default: "redshift"
+        required: true
```
emmyoop, line 43: If you drop them in the workflow_dispatch above, you cane drop these too.

.github/workflows/release-internal.yml
```diff
@@ -0,0 +1,64 @@
...
+  workflow_call:
```
emmyoop, line 24: When would this be called by another workflow?

.github/workflows/release-internal.yml
```diff
@@ -0,0 +1,64 @@
...
+    shell: bash
+
+env:
+  PYTHON_TARGET_VERSION: 3.11
+
+jobs:
+  invoke-reusable-workflow:
+    name: Build and Release Internally
+
+    uses: VersusFacit/dbt-release/.github/workflows/internal-archive-release.yml@main
```
emmyoop, line 56: Just noting this needs to be updated once the dbt-release PR gets merged.
```suggestion
    uses: dbt-labs/dbt-release/.github/workflows/internal-archive-release.yml@main
```

.github/workflows/release-internal.yml
```diff
@@ -0,0 +1,64 @@
...
+defaults:
+  run:
+    shell: bash
+
+env:
+  PYTHON_TARGET_VERSION: 3.11
+
+jobs:
+  invoke-reusable-workflow:
+    name: Build and Release Internally
```
emmyoop, line 54:
```suggestion
    name: "Build and Release Internally"
```

setup.py
```diff
@@ -50,7 +61,7 @@ def _plugin_version() -> str:
     install_requires=[
         "dbt-common>=0.1.0a1,<2.0",
         "dbt-adapters>=0.1.0a1,<2.0",
-        f"dbt-postgres~={_plugin_version()}",
+        f"dbt-postgres~={_plugin_version_trim()}",
```
emmyoop, line 64: @mikealfare can this be done without `~=`?

emmyoop (changes requested):
Looks good - mostly some questions/changes for the inputs.

.github/workflows/release-internal.yml
mikealfare: Why would we have a field that's both required and has a default? Wouldn't the default never be used?

.github/workflows/release-internal.yml
mikealfare: This is something we just added to other workflows because the test command used to be `dbt --version`, which no longer works on an adapter. I don't think it's necessarily a bad thing to keep this here until we settle on what this command should be.

setup.py
mikealfare: I would need to look at it, but I would think so. We would need to update the function to return a tuple, something along the lines of `"dbt-postgres>={},<{}".format(_plugin_version())` where `_plugin_version() == (1.8.0b1, 1.9)`.

.github/workflows/release-internal.yml
mikealfare: We've been very inconsistent in this aspect. I like less redundancy, e.g. not adding quotes around strings that yml inherently interprets as strings anyway. However, there are scenarios that require the quote, such as when there is a variable in the string `"Build ${{ inputs.version }}"`. So I agree that we should use quotes everywhere for consistency.

So my actual question, should we write this down in a "style guide" somewhere?

.github/workflows/release-internal.yml
VersusFacit: The default should be in text form upon dispatch, so it's a default argument, not a default value if nothing given.

.github/workflows/release-internal.yml
VersusFacit: Yeah I had a reason for this at some point but no longer. Good catch

.github/workflows/release-internal.yml
VersusFacit: Copied following a possible in-house convention. Happy to drop.

.github/workflows/release-internal.yml
emmyoop: Exactly.  Make it required so we're explicit.  But give it the default if we expect to be using main the majority of the time.

emmyoop (approved):
Approving as is for now with the understanding that more PRs will follow on after testing since none of this can be tested before the workflow gets merged.  Since this is a workflow that will be manually submitted there is little risk in this approach.

mikealfare:
Tests are failing here due to the `ddtrace` break. This is an unrelated issue. The primary motivation of merging this PR is to get a workflow in `main` that can be tested in a future PR. This PR represents partial, and on-going work.

PR #723, opened 2024-02-27
Add test

VersusFacit:
resolves #722 

Add a test after changing frozenset instantiation in Relation.

### Checklist

- [X] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [X] I have run this code in development and it appears to resolve the stated issue
- [X] This PR includes tests, or tests are not required/relevant for this PR
- [X] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc) or this PR has already received feedback and approval from Product or DX

PR #737, opened 2024-03-21
[Backport 1.7.latest] Fix the renamed relations code

mikealfare:
Backport c99c73bd0386e2641c825b8e36c5bc0229cd3258 from #723.

### DIFF
--- a/tests/unit/test_renamed_relations.py
+++ b/tests/unit/test_renamed_relations.py
@@ -1,5 +1,6 @@
+from dbt.contracts.relation import RelationType
+
 from dbt.adapters.redshift.relation import RedshiftRelation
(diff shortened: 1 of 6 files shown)


tests/unit/test_renamed_relations.py
```diff
@@ -0,0 +1,18 @@
+from dbt.contracts.relation import RelationType
```
mikealfare, line 1: This line was manually updated from `from dbt.adapters.contracts.relation import RelationType`. This class was moved during the decoupling of `dbt-core` and `dbt-adapters`, which happened in `1.8`.

martynydbt (approved):

PR #747, opened 2024-04-01
[Backport 1.7.latest] Test ddtrace dependency range (#733)

mikealfare:
Backport #733


mikealfare:
This is only needed for older versions.

PR #758, opened 2024-04-05
 Add docker release to the full release process for final releases

mikealfare:
### Problem

Publishing a Docker release is currently a manual process which can be forgotten.

### Solution

Add the Docker release to the release pipeline.

### Checklist


.github/workflows/release.yml
```diff
@@ -195,34 +184,38 @@ jobs:
...
       PYPI_API_TOKEN: ${{ secrets.PYPI_API_TOKEN }}
       TEST_PYPI_API_TOKEN: ${{ secrets.TEST_PYPI_API_TOKEN }}
 
+  docker-release:
+    name: "Docker Release"
+    if: ${{ !failure() && !cancelled() }}
+    needs: [bump-version-generate-changelog, build-test-package, github-release]
+    permissions:
+      packages: write
+    uses: dbt-labs/dbt-release/.github/workflows/release-docker.yml@docker-release
```
mikealfare, line 202: Point this back to main

docker/test.sh
```diff
@@ -0,0 +1,22 @@
+# - VERY rudimentary test script to run latest + specific branch image builds and test them all by running `--version`
+# TODO: create a real test suite
+
+clear \
```
VersusFacit, line 4: This entire thing could be cleaned up by just setting `-e` using `setopts` at the top and then not having to `&&` and `\` every line.

VersusFacit (dismissed):
One call for some refactoring but otherwise looking good.

docker/test.sh
mikealfare: lol, probably right, this is copy/pasta from the original and I don't think it was ever used. I wouldn't be against removing it entirely since I don't think it is used anywhere. And if it is, the check can be replaced with a one-liner.

docker/Dockerfile
```diff
@@ -0,0 +1,37 @@
+# this image gets published to GHCR for production use
+ARG py_version=3.10.7
```
emmyoop, line 2:
```suggestion
ARG py_version=3.11.2
```

mikealfare:
Test release using this branch: https://github.com/dbt-labs/dbt-redshift/actions/runs/8929496802

docker/test.sh
```diff
@@ -0,0 +1,19 @@
+# - VERY rudimentary test script to run latest + specific branch image builds and test them all by running `--version`
```
VersusFacit, line 1: Can probably nix

VersusFacit (approved):
Small nit, otherwise matches the PG PR

PR #774, opened 2024-04-18
Add integration tests for IAM User auth

### LINKED ISSUE #760: [Feature] Support IAM role and IAM user testing in CI

### Describe the feature

In order to be able to properly support redshift auth methods we need to have a functional test(s) that validate each mode works as expected given proper user input. 

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

mikealfare:
resolves #760

### Problem

We currently support IAM User auth but we don't currently have any tests that verify this.

### Solution

- add required infrastructure (users, roles, permissions, etc.) in AWS to support tests
- add tests for Database method auth
- add tests for IAM User method auth using a profile (give `redshift_connector` a profile name in your AWS config file)
- add tests for IAM User method auth using direct args (give `redshift_connector` the values that normally live in your AWS config file)
- update `test.env.example` so that contributors know what to provide to test this in their environment
- clean up unit tests since (I got lost trying to find the applicable tests since they were all in "test_redshift_adapter")

### Checklist


dbt/adapters/redshift/connections.py
```diff
@@ -208,14 +210,27 @@ def connect():
                 )
 
             def connect():
+                if profile := self.credentials.iam_profile:
```
mikealfare, line 213: Allow for both the file approach, and the direct approach.

tests/conftest.py
```diff
@@ -1,22 +1 @@
...
-
-# Import the functional fixtures as a plugin
-# Note: fixtures with session scope need to be local
-
 pytest_plugins = ["dbt.tests.fixtures.project"]
-
-
-# The profile dictionary, used to write out profiles.yml
-@pytest.fixture(scope="class")
-def dbt_profile_target():
```
mikealfare, line 12: This got moved within the `tests/functional` scope as live database credentials should never be needed for unit testing.

tests/functional/conftest.py
```diff
@@ -0,0 +1,18 @@
+import os
+
+import pytest
+
+
+# The profile dictionary, used to write out profiles.yml
+@pytest.fixture(scope="class")
+def dbt_profile_target():
```
mikealfare, line 8: This came from `tests/conftest.py`. See comment above.

dbt/adapters/redshift/connections.py
```diff
@@ -208,14 +210,22 @@ def connect():
                 )
 
             def connect():
+                if access_key_id := self.credentials.access_key_id:
```
mikealfare, line 213: Allow for both the file approach, and the direct approach.

colin-rogers-dbt:
Successful run with the new workflow: https://github.com/dbt-labs/dbt-redshift/actions/runs/8762024735/job/24112742150

dbt/adapters/redshift/connections.py
```diff
@@ -121,6 +121,8 @@ class RedshiftCredentials(Credentials):
     region: Optional[str] = None
     # opt-in by default per team deliberation on https://peps.python.org/pep-0249/#autocommit
     autocommit: Optional[bool] = True
+    access_key_id: Optional[str] = None
```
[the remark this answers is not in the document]
colin-rogers-dbt: great idea!

dbt/adapters/redshift/connections.py
colin-rogers-dbt: we should make sure to document this (even if it's intuitive) that providing an access_key_id overrides your local AWS cli setup

colin-rogers-dbt: also we should explicitly check/raise an exception for providing both a `secret_access_key` and a `access_key_id` (i.e. if you provide one then you need to provide both)

tests/conftest.py
colin-rogers-dbt: make sure to run the unit test workflow manually to validate we don't have some weird dependency on it

mikealfare: Agreed, but I'd almost rather fix that than move it back, or just move the tests into `functional`.

dbt/adapters/redshift/connections.py
mikealfare: > we should make sure to document this (even if it's intuitive) that providing an access_key_id overrides your local AWS cli setup

Good point, given we're effectively adding an auth method here, we need docs either way. But we need to explicitly call out the order in which `dbt-redshift` will look for creds.

> also we should explicitly check/raise an exception for providing both a secret_access_key and a access_key_id (i.e. if you provide one then you need to provide both)

I thought about doing this and didn't because python is a leap without looking language. However, I didn't think about what this looks like for the user. I'll add that in.

mikealfare: Along those lines, should we also add checks for providing both a profile and the explicit creds? Or simply document the order?

dbt/adapters/redshift/connections.py
```diff
@@ -174,60 +197,59 @@ def get_connect_method(self):
...
+            iam_kwargs.update(
+                {
+                    "access_key_id": self.credentials.access_key_id,
+                    "secret_access_key": self.credentials.secret_access_key,
+                }
+            )
 
         else:
-            raise FailedToConnectError("Invalid 'method' in profile: '{}'".format(method))
+            iam_kwargs.update(
```
mikealfare, line 228: The original logic flow allows for `iam_profile` to be empty and pass successfully. There's a unit test for it. So I check if `access_key_id` and `secret_access_key` are both provided, otherwise default to `iam_profile`. I'm not checking for the scenario where exactly one of `access_key_id` or `secret_access_key` are provided. This would lead to the user providing one of those, and then it tries to connect via the "profile, but no profile provided" method.

tests/conftest.py
mikealfare: These pass.

dbt/adapters/redshift/connections.py
```diff
@@ -142,14 +144,14 @@ def _connection_keys(self):
             "region",
             "sslmode",
             "region",
-            "iam_profile",
```
mikealfare, line 145: This was duplicated in the list.

dbt/adapters/redshift/connections.py
mikealfare: This was added

colin-rogers-dbt (approved):

PR #813, opened 2024-05-06
Mark flaky tests

### LINKED ISSUE #770: [Spike] Debug Redshift CI/CD failures

### Current Behavior

currently having a series flaky test in ci/cd (not appearing locally) 

### Expected Behavior

nightly ci/cd tests should pass ideally

### Steps To Reproduce

check alert-channel for previous runs of tests on various branches to track common cause and rerun

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt-core:
- dbt-redshift:
```

### Additional Context

_No response_

mikealfare:
resolves #770 

### Problem

We have a handful of flaky tests in `dbt-redshift` that routinely prevent CI from passing. It's not always the same test that fails. After a few runs, all tests have passed at least once, but it's rare that they all pass at the same time.

### Solution

- mark flaky tests using `pytest.mark` (e.g. `@pytest.mark.flaky`)
- update the existing integration tests to only run non-flaky tests
- add a new set of integration tests that only run flaky tests
  - re-running duration is greatly reduced
- reduce collisions from tests running at the same time
  - run these specifically after the non-flaky tests
  - run them in series according to GHA
  - run them in series according to pytest
- only require non-flaky tests
- only alert on failed non-flaky tests
- identify similar characteristics across flaky tests to troubleshoot the source

### Checklist


mikealfare:
The integration tests workflow is updated as part of this PR. Here's a run against this branch:
https://github.com/dbt-labs/dbt-redshift/actions/runs/8977125106

.github/workflows/integration.yml
```diff
@@ -53,85 +53,30 @@ defaults:
...
-      )
     runs-on: ${{ matrix.os }}
 
-    needs: test-metadata
-
     strategy:
       fail-fast: false
       max-parallel: 3
-      matrix: ${{ fromJSON(needs.test-metadata.outputs.matrix) }}
+      matrix:
```
mikealfare, line 69: This matrix is the equivalent of the non-PR matrix from `integration-test-matrix.js`, and is a superset of the PR matrix from `integration-test-matrix.js`. We are not losing coverage by moving to this config, and it's much easier to reference.

.github/workflows/integration.yml
```diff
@@ -53,85 +53,30 @@ defaults:
     shell: bash
 
 jobs:
-  # generate test metadata about what files changed and the testing matrix to use
-  test-metadata:
```
mikealfare, line 57: This job is no longer needed since we are using the inline matrix below.

.github/workflows/integration.yml
```diff
@@ -235,9 +179,96 @@ jobs:
       - uses: actions/upload-artifact@v3
         if: always()
         with:
-          name: integration_results_${{ matrix.python-version }}_${{ matrix.os }}_${{ matrix.adapter }}-${{ steps.date.outputs.date }}.csv
+          name: integration_results_${{ matrix.python-version }}_${{ matrix.os }}_redshift-${{ steps.date.outputs.date }}.csv
           path: integration_results.csv
 
+  test-flaky:
```
mikealfare, line 185: I skipped some steps from the non-flaky (original) version:
- artifacts are not uploaded
- AWS profile is not configured (none of the tests that would use this have been identified as flaky)

.github/workflows/main.yml
```diff
@@ -173,7 +173,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-12, windows-latest]
+        os: [ubuntu-latest, macos-12, windows-2022]
```
mikealfare, line 176: I updated this while troubleshooting some debug messages. While it wasn't the source, I figured I would leave this pinned to the current image that `windows-latest` points to since we did this for `macos-latest` recently.

tox.ini
```diff
@@ -21,6 +21,7 @@ passenv =
     REDSHIFT_TEST_*
     PYTEST_ADDOPTS
     DD_CIVISIBILITY_AGENTLESS_ENABLED
+    DD_INSTRUMENTATION_TELEMETRY_ENABLED
```
mikealfare, line 24: Reduces the amount of noise in logs for Windows CI tests.

.github/workflows/integration.yml
```diff
@@ -53,88 +53,33 @@ defaults:
...
-          persist-credentials: false
-
-      - name: Check out the repository (PR)
-        if: github.event_name == 'pull_request_target'
-        uses: actions/checkout@v3
-        with:
-          persist-credentials: false
-          ref: ${{ github.event.pull_request.head.sha }}
-
-      - name: Check if relevant files changed
```
colin-rogers-dbt, line 82: I think we still want the path filtering check right?

colin-rogers-dbt (commented):
Couple questions, really like the idea

.github/workflows/integration.yml
```diff
@@ -53,88 +53,33 @@ defaults:
...
-        github.event.pull_request.head.repo.full_name == github.repository ||
-        contains(github.event.pull_request.labels.*.name, 'ok to test')
-      )
     runs-on: ${{ matrix.os }}
 
-    needs: test-metadata
-
     strategy:
       fail-fast: false
-      max-parallel: 3
```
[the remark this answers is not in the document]
mikealfare: It didn't seem to affect it, but we could put it back. I did that because windows-latest wound up last, but takes the longest. I can put the limit back, but build the matrix in a way that puts windows in the first batch.

.github/workflows/integration.yml
```diff
@@ -53,88 +53,33 @@ defaults:
...
-      matrix: ${{ fromJSON(needs.test-metadata.outputs.matrix) }}
+      matrix:
+        os: [ubuntu-latest]
```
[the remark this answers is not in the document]
mikealfare: Yeah, it crossed my mind. I justified it for windows because I was actually updating it as part of troubleshooting. But I'd like to do both, so I will.

.github/workflows/integration.yml
```diff
@@ -53,88 +53,33 @@ defaults:
...
+          - os: windows-2022
+            python-version: 3.8
 
     env:
-      TOXENV: integration-${{ matrix.adapter }}
+      TOXENV: integration-redshift
       PYTEST_ADDOPTS: "-v --color=yes -n4 --csv integration_results.csv"
       DBT_INVOCATION_ENV: github-actions
       DD_CIVISIBILITY_AGENTLESS_ENABLED: true
+      DD_INSTRUMENTATION_TELEMETRY_ENABLED: false
```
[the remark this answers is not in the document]
mikealfare: This removes the extra debug logging that we're seeing for windows. Honestly, it doesn't really matter in CI, and can be removed.

.github/workflows/integration.yml
mikealfare: I could work at adding that back in. Leaving it out just means that we run integration tests on all PRs.

mikealfare: This actually doesn't get used. It gets added to the environment variables for the next step, but doesn't get called. It also doesn't get returned as an output, so it's a no op.

mikealfare:
Here's the latest integration tests run against this branch:
https://github.com/dbt-labs/dbt-redshift/actions/runs/9011428584

colin-rogers-dbt (approved):

PR #836, opened 2024-05-31
Remove group_federation for serverless

### LINKED ISSUE #835: [Bug] dbt-redshift should not use `group_federation=true` when connecting with Serverless.
Group_federation parameter is not required with Serverless and makes the redshift connector call `getclustercredentialswithIAM` instead of `get_credentials` - the former doesn't exist for Serverless.

Fleid:
See this convo: https://dbt-labs.slack.com/archives/C04FN4114AD/p1717037952388479

Resolves #835 835

This has no test :/
But I confirmed the connection now hits redshift instead or returning:
```
  >Database Error
  'RedshiftServerless' object has no attribute 'get_cluster_credentials_with_iam'
```


mikealfare (dismissed):

mikealfare:
This [unit test](https://github.com/dbt-labs/dbt-redshift/blob/cef201f95677dd2dec5f0875ea67a258f4d3a6b0/tests/unit/test_auth_method.py#L398-L458) should be updated for both the serverless and cluster_id scenarios.

dbt/adapters/redshift/connections.py
```diff
@@ -234,10 +234,15 @@ def _iam_user_kwargs(self) -> Dict[str, Any]:
...
-        kwargs.update(
-            group_federation=True,
-            db_user=None,
-        )
+
+        # It's a role, we're ignoring the user
+        kwargs.update(db_user=None)
+
+        # Serverless shouldn't get group_federation, Provisoned clusters should
+        if "serverless" in self.credentials.host:
```
mikealfare, line 242: It's still kind of wild that this is the right way to do this.

.changes/unreleased/Fixes-20240531-113620.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Fixes
+body: Remove group_federation when using serverless for iam authentication
```
mikealfare, line 2: I could take it or leave it, but this might read better to an end user as:
```
Support IAM Role authentication for Redshift Serverless
```
What's here is _how_ we did it versus _what_ the user gets.

mikealfare (approved):

.changes/unreleased/Fixes-20240531-113620.yaml
Fleid: Done!
You would think I know that...

dbt/adapters/redshift/connections.py
Fleid: IKR

PR #844, opened 2024-06-11
Pin to 2.1.1 and below to fix the cursor error.

VersusFacit:
### Problem
We want to fold in the fix from [this patched issue on redshift connector](https://github.com/aws/amazon-redshift-python-driver/issues/204).

### Solution
Up the pin. soft pinned since either patch version in `2.1` should work

### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -64,7 +64,7 @@ setup(
         f"dbt-postgres~={_plugin_version_trim()}",
         # dbt-redshift depends deeply on this package. it does not follow SemVer, therefore there have been breaking changes in previous patch releases
         # Pin to the patch or minor version, and bump in each new minor version of dbt-redshift.
-        "redshift-connector<2.0.918,>=2.0.913,!=2.0.914",
+        "redshift-connector<=2.1.1",
         # add dbt-core to ensure backwards compatibility of installation, this is not a functional dependency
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -64,7 +64,7 @@ def _plugin_version_trim() -> str:
         f"dbt-postgres~={_plugin_version_trim()}",
         # dbt-redshift depends deeply on this package. it does not follow SemVer, therefore there have been breaking changes in previous patch releases
         # Pin to the patch or minor version, and bump in each new minor version of dbt-redshift.
-        "redshift-connector<2.0.918,>=2.0.913,!=2.0.914",
+        "redshift-connector<2.1.1,>=2.0.913,!=2.0.914",
```
mikealfare, line 67: Is there a reason to pin less than a specific patch? I know we've been broken in the past, but if this package is indicating that they're going to start using minors, then I would think we pin to 2.1, not 2.1.x.

mikealfare (approved):
I left a comment, but I'm fine with the PR either way.

setup.py
colin-rogers-dbt: we should follow up with them and see if that's actually what's happening here or if they just decided to do a minor version bump

PR #875, opened 2024-07-19
Remove `freezegun` as a test dependency since it is not used

mikealfare:
### Problem

We list `freezegun` as a test dependency, but it's not used anywhere. It just winds up generating `dependabot` PRs.

### Solution

Remove the dependency. It was left over from the monorepo but does not apply here.

### Checklist


VersusFacit (approved):
assuming you get tests to pass

PR #885, opened 2024-07-22
Adap 821/unit tests infer wrong datatype for None values in fixtures

### LINKED ISSUE #821: [Bug] Unit test infers the wrong data type when the first record has some NULL values

- [X] I believe this is a new bug in dbt-core

### Current Behavior

```yml
unit_tests:
  - name: test_user_events_combined
    model: int_mycfg__user_events_combined
    given:
      - input: ref('stg_mycfg__user_events')
        rows:
          - {
              created_at: "2024-01-01 10:01:00",
              user_id: 1,
              is_closed: null, # <--- if we change both of these null values to true/false, the error goes away.
              is_validated: null, # <--- if we change both of these null values to true/false, the error goes away.
              user_action: created,
            }
          - {
              created_at: "2024-01-01 10:02:00",
              user_id: 1,
              is_closed: null,
              is_validated: true,
              user_action: validated,
            }
      - input: ref('stg_mycfg__user_snapshot')
        rows: []
    expect:
      rows:
        - { user_id: 1, is_closed: true, is_validated: true }
        - { user_id: 2, is_closed: null, is_validated: true }
        - { user_id: 3, is_closed: true, is_validated: null }
        - { user_id: 4, is_closed: null, is_validated: null }
        - { user_id: 5, is_closed: false, is_validated: true }
        - { user_id: 6, is_closed: true, is_validated: true }
```

When you run the test you will see error like this.

image

And here is the logs:

image

_Input build_

image

_Expected result build_

Interestingly, the incorrect casting happen in the `dbt_internal_unit_test_expected` cte but not the input cte.

---

Our current workaround is to use sql to cast the correct type, eg.

```yml
select
    '2024-01-01 10:01:00' as created_at,
    1 as user_id,
    null::boolean as is_closed,
    null::boolean as is_validated,
    'created' as user_action
union all
select '2024-01-01 10:02:00', 1, null, true, 'validated'
```

### Expected Behavior

The test runner should be able to guess the data type correctly.

### Steps To Reproduce

1. Provide unit test data with some null values in the first record.
2. Run the test
3. ERROR!

### Relevant log output

image

_Input build_

image

_Expected result build_

Interestingly, the incorrect casting happen in the `dbt_internal_unit_test_expected` cte but not the input cte.

### Environment

```markdown
- OS: Ubuntu 24.04 WSL
- Python: 3.12.2
- dbt: 1.8.0-rc1
```

redshift

### Additional Context

_No response_

VersusFacit:
resolves #821 

Note this Pr depends on the [base adapter PR here](https://github.com/dbt-labs/dbt-adapters/pull/270).

### Problem

Take 2 in response to #860 

### Solution

Redshift provides a macro that will check for the bad row scenario and throw a more helpful error to the user than the existing type mismatch bug that caused this error in the first place.

### Checklist

### DIFF
--- a/dbt/include/redshift/macros/adapters/unit_testing.sql
+++ b/dbt/include/redshift/macros/adapters/unit_testing.sql
@@ -1,0 +1,16 @@
+{% macro redshift__validate_fixture_rows(rows, row_number) %}
+  {%- if items is not none and items|length > 0 -%}
+    {%- set some_value_is_none = false -%}
+    {%- set row = items[0] -%}
+    {%- for key, value in row.items() -%}
+      {%- if value is none -%}
+        {%- set some_value_is_none = true -%}
+      {%- endif -%}
+    {%- endfor -%}
+
+    {%- if some_value_is_none -%}
+      {%- set fixture_name = "expected output" if model.resource_type == 'unit_test' else ("'" ~ model.name ~ "'") -%}
+      {{ exceptions.raise_compiler_error("Unit test fixture " ~ fixture_name ~ " in " ~ model.name ~ " does not have any row free of null values, which may cause type mismatch errors during unit test execution.") }}
+    {%- endif -%}
+  {%- endif -%}
+{%- endmacro -%}
--- a/dbt/include/redshift/macros/adapters/unit_testing.sql
+++ b/dbt/include/redshift/macros/adapters/unit_testing.sql
@@ -1,16 +1,11 @@
-{% macro redshift__validate_fixture_rows(rows, row_number) %}
-  {%- if items is not none and items|length > 0 -%}
-    {%- set some_value_is_none = false -%}
-    {%- set row = items[0] -%}
+{%- macro redshift__validate_fixture_rows(rows, row_number) -%}
+  {%- if rows is not none and rows|length > 0 -%}
+    {%- set row = rows[0] -%}
     {%- for key, value in row.items() -%}
       {%- if value is none -%}
-        {%- set some_value_is_none = true -%}
+          {%- set fixture_name = "expected output" if model.resource_type == 'unit_test' else ("'" ~ model.name ~ "'") -%}
+          {{ exceptions.raise_compiler_error("Unit test fixture " ~ fixture_name ~ " in " ~ model.name ~ " does not have any row free of null values, which may cause type mismatch errors during unit test execution.") }}
       {%- endif -%}
     {%- endfor -%}
-
-    {%- if some_value_is_none -%}
-      {%- set fixture_name = "expected output" if model.resource_type == 'unit_test' else ("'" ~ model.name ~ "'") -%}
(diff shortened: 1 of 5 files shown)


dbt/include/redshift/macros/adapters/unit_testing.sql
```diff
@@ -0,0 +1,11 @@
+{%- macro redshift__validate_fixture_rows(rows, row_number) -%}
+  {%- if rows is not none and rows|length > 0 -%}
+    {%- set row = rows[0] -%}
+    {%- for key, value in row.items() -%}
+      {%- if value is none -%}
+          {%- set fixture_name = "expected output" if model.resource_type == 'unit_test' else ("'" ~ model.name ~ "'") -%}
+          {{ exceptions.raise_compiler_error("Unit test fixture " ~ fixture_name ~ " in " ~ model.name ~ " does not have any row free of null values, which may cause type mismatch errors during unit test execution.") }}
```
mikealfare, line 7: Did we still want to raise an error here? I remember we talked about potentially just a warning instead of an error, but that could be old information.

dev-requirements.txt
```diff
@@ -1,5 +1,5 @@
 # install latest changes in dbt-core + dbt-postgres
-git+https://github.com/dbt-labs/dbt-adapters.git
+git+https://github.com/dbt-labs/dbt-adapters.git@ADAP-821/support_for_redshift_821
```
mikealfare, line 2: Friendly reminder to revert prior to merging.

tests/functional/adapter/unit_testing/test_unit_testing.py
mikealfare, line 1: I know we wanted to do this in a way that didn't impact other adapters. But would it be possible to take the generic bits here and put them in `dbt-tests-adapter`? We could simply not implement the tests in the other adapters. But if we did want to implement this in other adapters, then we could reuse those tests. Or is the idea that these specifically need to be overwritten for `dbt-redshift`?

VersusFacit: This is specifically a redshift problem. The other adapters don't rely on this behavior, so I saw no reason to duplicate these tests elsewhere.

dbt/include/redshift/macros/adapters/unit_testing.sql
VersusFacit: Okay, so Michelle and I aligned on this, and I'll see what you think!

Basically there is a special case that we're still throwing an error in Redshift (there is no all-`None` value row to promote to the first row of the fixture under the hood). 

This special case throws a cryptic error for redshift users specifically.

So now we throw a helpful error that documents this behavior both for user in logs and us in code.

I proposed that this does not introduce new behavior because this was already an error and will continue to be.

The logic change made in core prevents a large percentage of the time this exception would be thrown.

mikealfare (approved):
Approved, conditional on reverting `dev-requirements.txt` post merge of https://github.com/dbt-labs/dbt-adapters/pull/270.

PR #910, opened 2024-09-11
Isolating distribution testing

### LINKED ISSUE #874: Update the build check workflow

### Current Behavior

a

### Expected Behavior

a

### Steps To Reproduce

a

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt-core:
- dbt-redshift:
```

### Additional Context

_No response_

leahwicz:
resolves #874  

### Problem
We validate wheel and source distributions in the same environment during the build validation step. This led to issues documented here https://github.com/dbt-labs/dbt-core/issues/10465. 

### Solution
This is the same solution as what we did in `dbt-snowflake` first: https://github.com/dbt-labs/dbt-snowflake/pull/1161
This is to bring parity to the other repos as well.

This is a test run of the Action running with the installation and testing of wheels and source distributions split into separate Actions Jobs that isolate them from one another: https://github.com/leahwicz/dbt-redshift/actions/runs/10821351505

### Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -48,7 +48,7 @@ jobs:
           persist-credentials: false
 
       - name: Set up Python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: '3.8'
 
@@ -83,7 +83,7 @@ jobs:
           persist-credentials: false
 
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
 
@@ -125,7 +125,7 @@ jobs:
           persist-credentials: false
 
       - name: Set up Python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: '3.8'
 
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -48,7 +48,7 @@ jobs:
           persist-credentials: false
 
       - name: Set up Python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
```
leahwicz, line 51: This repo is behind on some of the versions of Actions that are updated in the other repos. I figured I would update them while changing this file to match what we have everywhere else. The run still passes without issues on the upgrade

.github/workflows/main.yml
```diff
@@ -173,41 +173,34 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-22.04, macos-12, windows-2022]
+        os: [ubuntu-latest, macos-12, windows-latest]
```
leahwicz, line 176: Same here, I remember wanting to keep on the latest images (this isn't the case for Mac, not sure why). So the other adapter repos use `latest` instead of a hard coded version so updating this to keep parity as well. I can revert this if it isn't wanted

mikealfare: We determined that it's better to hard pin and then upgrade as we can. The primary issue is that the latest tag eventually bumps and the new image may not have all of the python versions installed. We ran into this on macos in particular, which is probably what we used to pin that in the past.

PR #919, opened 2024-09-27
[Bug] Fix scenario where dbt attempts to add existing columns to relations when using the SDK for column metadata

### LINKED ISSUE #914: [Bug] dbt build fails when flag restrict_direct_pg_catalog_access set True

### Current Behavior

When I switch dbt cloud to versionless and I build with no flags set then the build works, but with Warnings to set the flag `restrict_direct_pg_catalog_access`

When I set the flag `restrict_direct_pg_catalog_access` to True the build fails, with a number of tables (probably around 10%) Failing
```
flags:
  restrict_direct_pg_catalog_access: True
```

Mostly with one of these Errors:
```
Redshift adapter: Redshift error: column "id" of relation "<RELATION_NAME>" already exists
Redshift adapter: Redshift error: syntax error at or near ")" in context 
```

Setting it to False or removing the flag means the build works again.

### Expected Behavior

The build to succeed with no Warnings.

### Steps To Reproduce

In dbt cloud with the project version set to `versionless`
Set the following flag in `dbt_project.yml`
```
flags:
  restrict_direct_pg_catalog_access: True
```
Run `dbt build`

### Relevant log output

_No response_

### Environment

```markdown
- OS: dbt cloud
- Python: dbt cloud
- dbt-core: versionless - 'Running latest version'
- dbt-redshift: redshift=1.9.0-post5+04bd2c07f526e226c80f6527e8b0d49033bbd639
```

### Additional Context

image
In the above

1. Added the flag set to True in the dbt_project.yml
2. Removed the flag completely
3. Added the flag back in set to False (to check it wasn't due to mis-typed config)
4. Change the flag back to True... you know just to check again.

mikealfare:
resolves #914

### Problem

When using the `redshift_connector` SDK to collect metadata about a relation, `dbt-redshift` attempts to add a column to a model which already exists on the model. This does not occur when using the legacy metadata query.

### Solution

Still investigating, but it appears this could be related to case sensitivity and/or quoting. Changing a column from unquoted to quoted on an incremental model results in the error that the user reported. However, this error also occurs when using the legacy metadata query.

### Checklist


tests/functional/utils.py
```diff
@@ -0,0 +1,8 @@
+from dbt.tests.util import get_model_file, relation_from_name, set_model_file
+
+
+def update_model(project, name: str, model: str) -> str:
```
mikealfare, line 4: This was copy/pasted from other adapters. We should move it to `dbt-tests-adapter` eventually, but that would complicate this PR. It should be done as a separate exercise.

dbt/adapters/redshift/impl.py
```diff
@@ -73,7 +73,7 @@ def _behavior_flags(self) -> List[BehaviorFlag]:
         return [
             {
                 "name": "restrict_direct_pg_catalog_access",
-                "default": False,
+                "default": True,
```
mikealfare, line 76: This allows us to run all tests as if the flag were enabled without adjusting the tests. This should be reverted prior to merging.

tests/functional/columns_in_relation_tests/test_columns_in_relation.py
```diff
@@ -28,6 +24,10 @@ def test_columns_in_relation(self, project, expected_columns):
         )
         with project.adapter.connection_named("_test"):
             actual_columns = project.adapter.get_columns_in_relation(my_relation)
+        expected_columns = [
```
mikealfare, line 27: Since we are aliasing dtypes back to the original names, we don't need to specify the expected column dtypes based on whether the flag is enabled.

mikealfare:
We elected to remove this functionality until we can resolve the discrepancy between our custom query and `redshift_connector`'s `get_columns` method. This is no longer relevant.

PR #931, opened 2024-10-16
Drop support for Python 3.8

mikealfare:
### Problem

Python 3.8 reached end of life on 10/7/2024.

### Solution

Drop support for Python 3.8. Update tests to no longer run on Python 3.8.

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc) or this PR has already received feedback and approval from Product or DX

PR #938, opened 2024-10-23
[Backport 1.4.latest] Drop support for Python 3.8 #2

mikealfare:
Backport #931 addition

The python version was specified in `.pre-commit-config.yml`, which was missed during the backport as this is not the case in newer versions.


VersusFacit (approved):
I changed the base from `main`. I am paranoid, so PLEASE double check it's right now

mikealfare:
The above tests passed. The failures are due to `pytest` not finding anything and returning an exit code of 5 instead of 0, which was not handled in the workflows for older versions.

PR #942, opened 2024-11-06
Test `macos-14` on Python 3.10 instead of Python 3.9 due to missing `psycopg2-binary`

mikealfare:
### Problem

`psycopg2-binary` did not publish a wheel for Python 3.9 on `macos-14`.

### Solution

Test `macos-14` against Python 3.10.

### Checklist


mikealfare:
The failed CI run is expected; that is the issue this PR resolves, but the check runs in the target context (`main`), so we don't see it working here.

PR #954, opened 2024-11-20
Fix macos14 tests on 1.5

VersusFacit:
resolves #

### Problem

1.4 tests will fail on macos until we fix this.

### Solution

Go to a version with a psycopg2 binary.

### Checklist

### DIFF
--- a/.github/scripts/integration-test-matrix.js
+++ b/.github/scripts/integration-test-matrix.js
@@ -1,6 +1,6 @@
 module.exports = ({ context }) => {
-  const defaultPythonVersion = "3.9";
-  const supportedPythonVersions = ["3.9", "3.10", "3.11"];
+  const defaultPythonVersion = "3.10";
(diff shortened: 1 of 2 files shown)


VersusFacit:
Judging from this manual CI run, we need to use parens around these....

https://github.com/dbt-labs/dbt-redshift/actions/runs/11927014503

Are PR checks still using the already-checked-in version? If so, this would explain for why they failed here but passed in the linked CI run

.github/scripts/integration-test-matrix.js
```diff
@@ -1,6 +1,6 @@
 module.exports = ({ context }) => {
-  const defaultPythonVersion = "3.9";
-  const supportedPythonVersions = ["3.9", "3.10", "3.11"];
+  const defaultPythonVersion = "3.10";
+  const supportedPythonVersions = ["3.9", "3.10", "3.11", "3.12"];
```
mikealfare, line 3: We don't support 3.12 on this version.

mikealfare (approved):

PR #985, opened 2025-01-30
Combine the build and check build jobs into one job

mikealfare:
Combine the build and check build jobs into one job. This allows the artifacts from each scenario in the matrix to be tested and avoids the need for uploading and downloading the built artifacts.

This needs to be backported to 1.4 as all versions are currently failing this check.


mikealfare:
We will not merge this into `main` since this branch is effectively stale now.

